PR #5, opened 2018-05-30
Custom transaction name, et al

bmacnaughton:
- Custom transaction name (oboe_http_span only).
- Remove amqp from tests.
- Create log.patching() for patch-time issues.
- Domain prefix.
- Improve express testing.
- Add tests for custom transaction name, domain prefix.
- add readyToSample()

Custom transaction names and domain prefixes are implemented for http-initiated traces.

## Custom Transaction Names

Express sets the default transaction name:
\``express.${func.name||'anonymous'}.${req.method}${req.route.path}`\`

where `func` is the function handling the express route and `req` is the http request object.

There are two ways to set a custom transaction name - request a callback when a custom name is needed and making an explicit call to set the transaction name. The first is most useful when setting a custom name for all, or most, routes. The second is commonly used when only a small subset of routes get custom names.

`ao.setCustomTxNameFunction('express', callback)` sets a function for the 'express' probe that will call `callback` when a custom name is needed. The express probe will receive two arguments, the http `req` and `res` objects as context for generating the custom name.

`ao.setCustomTxName(req, res, name)` needs to be reworked to accommodate non-http initiated traces.

## Domain Prefix

Setting the domain prefix is the config file property `domainPrefix`. If set it will be used as a prefix for transaction names, custom or not.

The logic for setting the domain prefix is as follows.

```
  let h = req.headers
  let s = req.socket || {}
  let prefix = h && h['x-forwarded-host'] || h['host'] || ''
  if (s.localPort !== 80 && s.localPort !== 443 && prefix !== '') {
    prefix = prefix + ':' + localPort
  }
  return prefix
```

## Note

This requires an unreleased version of appoptics-bindings and appropriate changes to package.json.


test/probes/express.test.js
```diff
@@ -88,41 +195,105 @@ describe('probes.express ' + pkg.version, function () {
...
+    } finally {
+      ao.cfg.domainPrefix = false
+    }
+  })
+
+  it('should handle an error in the custom name function', function (done) {
+    function custom (req, res) {
+      throw new Error('I am a bad function')
+    }
+    customTransactionName(undefined, done)
```
prettycoder: should this be
 `customTransactionName(**custom**, done)`

bmacnaughton: Yes, it should be. Thanks for catching this.

PR #8, opened 2018-06-06
rc1

bmacnaughton:
Custom transaction name [AO-7777]
- Use final transaction name on exit event.
- depends on appoptics-bindings@5 [AO-7992]
- domain prefix [AO-8104]
- default TransactionName for custom spans is 'custom-' prefix to the span name.

Bug fixes
- don't duplicate domain port if already in header
- never prefix domain for non-http traces [AO-8104]
- bind http emitters from active CLS context. [AO-8454]

General
- remove log.force functions and concept [AO-8224]
- add TOKEN_BUCKET env vars to env.sh
- move docs to jsdoc via jsdoc2md [AO-7403]
- add more examples to SDK/API documentation
- README links to github for "adding your own spans" docs.
- add tests in appoptics-apm and todo application

Examples

Using custom transaction name with SDK `startOrContinueTrace` pass an options object with the property `customTxName` with a value of either a string (use that name) or a function that returns the string name to be used.

```
  ao.startOrContinueTrace(
    null,
    'sync-span-name',
    functionToRun,
    // note - no context is provided for the customTxName function. If
    // context is required the caller should wrap the function in a closure.
    {customTxName: customNameFunction}
  )
```

To set a custom transaction name for autoinstrumented code, set a callback function for that instrumented package, e.g., `express` (the initial package supporting custom transaction names). The signature of callbacks is dependent on the specific package being instrumented but HTTP-based callbacks receive the `req` and `res` objects typically passed to callbacks.

```
ao.setCustomTxNameFunction('express', customNameFunction)

function customNameFunction (req, res) {
  // make a custom transaction name. the express default is
  // 'express.' + func.name + '.' + req.method + req.route.path
  // where func.name is the name of the function handling the route or 'anonymous'
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ npm install --save appoptics-apm@latest
 
 ## Adding Your Own Spans
 
-Documentation coming soon!
+Our GitHub repository hosts an [overview](https://github.com/appoptics/appoptics-apm-node/blob/master/guides/instrumenting-a-module.md) and a [complete API reference](https://github.com/appoptics/appoptics-apm-node/blob/master/guides/api.md).
 
 ## Support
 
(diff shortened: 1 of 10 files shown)


bmacnaughton:
index.js, span.js, probes/http.js, and custom.test.js are the primary files to be reviewed.

api.md should be skipped - it is generated from the docs in the files.

PR #23, opened 2018-09-06
Add support for Koa 2's new Koa constructor

josephschmitt:
Need to return a classic function so it can be called with the
`new` keyword, just as Koa is.

### DIFF
--- a/lib/probes/koa.js
+++ b/lib/probes/koa.js
@@ -5,11 +5,11 @@ const Span = require('../span')
 const ao = require('../')
 const conf = ao.probes.koa
 
-module.exports = koa => {
-  return () => {
-    const app = koa()
-    wrapApp(app)
-    return app
+module.exports = Koa => {
+  return function () {
+    const app = new Koa();
+    wrapApp(app);
+    return app;
   }
 }
 
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "gulp-istanbul": "^0.10.0",
     "gulp-mocha": "^3.0.1",
     "hapi": "^13.5.3",
-    "koa": "^1.6.0",
+    "koa": "^2.0.0",
     "koa-resource-router": "^0.4.0",
     "koa-route": "^2.4.2",
     "koa-router": "^5.4.2",


josephschmitt:
@bmacnaughton hey there. The big change here is using the `new` keyword. Koa 2 exports an ES5 `class` instead of a function. Without updating this to do `app = new Koa` instead of `app = koa()`, any application that requires this library and triggers the koa probe will throw an error on startup. I returned a function for backwards compatibility, since calling `new Koa` should work for both the Koa 2 class, and the Koa 1 function.

josephschmitt:
@bmacnaughton feel free to ignore this PR. I submitted it after talking to someone on your end about Koa 2 support and merely dropped this as a fewest-number-of-lines-changed solution to trying to get my Koa 2 app to start without errors. Let me know if I can help with getting a Koa 2 compatible version out and tested.

bmacnaughton:
yes, calling `new koa()` works for both v1 and v2. but the arrow function to classic function isn't important (until we target babel at a javascript version that supports arrow functions) - that's what i was trying to understand.

thanks for your efforts on this - we support a number of packages and i rarely know them as well as the people that use them day-in and day-out. this was helpful and will prevent any future problems - much appreciated.

i'm going to fix the issues and am working on rolling this change into a bigger release. please let me know if you run into any other problems.

bmacnaughton:
@josephschmitt the koa@2 version tests fine right now - it warns about deprecation of the generator-style chaining but still supports it.

i'm presuming you're using promises though. do you use bluebird, native promises, q, or another?

and, if you have spare time, looking at `test/probes/koa.js` and providing some insight into what testing would be more helpful (and how to merge in support for promises - i don't want to remove support for version 1 yet) would be quite valuable. the actual test driver is `test/probes/koa.test.js` but it calls the previous file for most of the heavy lifting with assistance from `test/helper.js`

the reason i'm preferring promise (as opposed to async/await - same thing though) is that promises don't cause the javascript engine to blow up for versions before 7.6.

if you'd want to do a google chat or something to go over testing i'd be happy to do that too.

bmacnaughton:
I'm going to get a release out today that supports koa2.

PR #31, opened 2018-10-18
Cleanse service name [AO-10253, AO-10347]

bmacnaughton:
- cleans service per https://swicloud.atlassian.net/browse/AO-10242
- sidesteps oboe validation by deleting/restoring APPOPTICS_SERVICE_KEY env var, if present
  - eliminates ambiguous oboe logs (overriding default using env var)
- tested using old version of oboe that doesn't transform bad service names.
  - oboe log 'client id' value verifies transformed name without override log message.
- fixes undefined FilePath error [AO-10347]

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -13,9 +13,9 @@ if (global[aoOnce]) {
   module.exports.loggers.warn('appoptics-apm is being executed more than once')
 } else {
 
-// disable eslint's indent it doesn't complain because everything in the else
-// (all of the file when it's required the first time) isn't indented.
 /* eslint-disable indent */
+// disable eslint's indent so it doesn't complain because everything in the else
+// (all of the file when it's required the first time) isn't indented.
 
 /**
  * @class ao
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -236,15 +236,36 @@ if (!config.ignoreConflicts) {
 // if the service key is defined in the environment then use that. if
 // not see if it is defined in the config file.
 //
-let serviceKey = env.APPOPTICS_SERVICE_KEY || config.serviceKey
-exports.serviceKey = serviceKey
+const environmentKey = process.env.APPOPTICS_SERVICE_KEY
```
prettycoder: Do you need `process.` when you defined `const env = process.env` above?

lib/index.js
```diff
@@ -310,7 +331,7 @@ function clsCheck (msg) {
 
 exports.clsCheck = clsCheck
 
-if (!serviceKey) {
+if (!cleansedKey) {
```
prettycoder: Would it make sense to move this closer to line 268

lib/index.js
```diff
@@ -717,7 +740,13 @@ if (!enabled) {
...
     options.hostnameAlias = exports.cfg.hostnameAlias
   }
-  bindings.oboeInit(serviceKey, options)
+
+  // delete the environment variable, init oboe, and restore it. this is
+  // done because oboe will prefer the environment variable to anything
+  // specified here.
+  delete env.APPOPTICS_SERVICE_KEY
+  bindings.oboeInit(cleansedKey, options)
+  env.APPOPTICS_SERVICE_KEY = environmentKey
```
prettycoder: Is there any imaginable scenario, where a service with a different config.serviceKey would be started from this env? If so, the env var should only be set if that is where the service_name was read from in the first place.

lib/index.js
bmacnaughton: I don't think so. The reason it's there is that I had a bug (I'd forgotten to change the parameter in the call to oboeInit()) and I added that on the wild chance that process itself used a proxy. I'm pretty sure I can change it back but testing this (and making sure it's not oboe changing the value) is a bit painful, so I just left it.

nice eye!

lib/index.js
bmacnaughton: It make sense to do something with it. It can issue a redundant message now, e.g., bad trace-mode, can't load bindings, etc. so the disabling might not be true (it could already be disabled). i'll have to think about what to do because it should either move up (and other potential messages check it) or stay where it is but modify the message and/or logic to be consistent.

prettycoder:
all questions answered. looks good :)

cheempz:
Jumping in with one more question: if the agent logs out the service key in a message, we'll need to ensure masking of the token part -- don't want to have the possibly sensitive API token written to file somewhere.  Example: https://github.com/appoptics/appoptics-apm-go/pull/83/files#diff-a2727f1742c6b1077d515d3a24a8fe7dL129

PR #37, opened 2018-11-07
5.9.0 [AO-9230, AO-9720, AO-10001, AO-10300, AO-10541]

bmacnaughton:
Fix CLS-not-active errors: amqplib, levelup.
Fix lost-context errors: cassandra-driver, mysql.
Fix ao.bindEmitter - now returns possibly bound function.
Move benchmark files from test/ to bench/.
Create common test require file.
Prevent simultaneous test collisions with AO_IX: mysql, cassandra-driver. mongodb-core, pg.

propagate .native functions to patched realpath and realpathSync for esm.

override traceqa/mongo-set REPLSETHOST with "localhost"


bmacnaughton:
main files to review are all in `lib/probes/` with extension `.js`

`fs.js` - help esm package avoid core dump. we were hiding functions that were node internal.

the following files either caused an error that a KV item was invalid (undefined) or lost context due to the patching not being quite correct.

`amqplib`
`cassandra-driver`
`levelup`
`mysql`
`redis`

lib/index.js
```diff
@@ -670,7 +670,8 @@ exports.bindEmitter = function (em) {
       // last event has been setup when the http instrumentation binds the
       // events but there must be CLS context.
       if (exports.tracing || (clsCheck() && (em.headers && em.socket))) {
-        return exports.requestStore.bindEmitter(em)
+        exports.requestStore.bindEmitter(em)
```
prettycoder: Question: Is this cosmetic or does it change the timing of when the bind happens?

lib/probes/amqplib.js
```diff
@@ -167,6 +168,8 @@ function patchConsume (proto) {
         () => cb.call(this, msg),
         conf
       )
-    }), options)
+    }
+    const f = ao.lastEvent ? ao.bind(consumePatched) : consumePatched
```
prettycoder: What kind of situation results in no lastEvent?
I was expecting an additional test to cover it, maybe I'm not looking in the right place.

lib/probes/cassandra-driver.js
```diff
@@ -85,8 +83,10 @@ module.exports = function (cassandra) {
     patchSendOnConnection(RequestHandler, sendOnConnection, connection)
   }
 
-  if (cassandra.Client && cassandra.Client.prototype) {
+  if (cassandra.Client && cassandra.Client.prototype && false) {
```
prettycoder: `&& false` temporary?

lib/index.js
bmacnaughton: requestStore.bindEmitter() does not return the bound emitter; it just wraps it, so the return value was incorrect previously.

lib/probes/amqplib.js
bmacnaughton: The primary place I see it in users programs is during startup. Because most of what I deal with are http servers it is generally the http request that initiates the tracing. Various programs (databases and queues, etc.) tend to set things up before the http server starts listening to requests.

`ao.lastEvent` is basically a `we are tracing` check.

lib/probes/cassandra-driver.js
bmacnaughton: you found the bug i just fixed! the tests found it. yes, it was temporary (or supposed to be!) in order to test the `logMissing()` function i'd added. all probes need to log when they can't find something they need to patch. previously i'd just hardcoded a separate message for each error but it became a pattern. so rather than duplicate the code and have many unique messages `index.js` now has a `makeLogMissing()` function that returns a function that can be used by each probe.

lib/probes/tedious.js
```diff
@@ -5,22 +5,47 @@ const ao = require('..')
...
+    }
+  }
+  if (!f) {
+    logMissing('connection.[createDebug, defaultConfig]()')
+    return
+  }
+
+  shimmer.wrap(connection, f, fn => function () {
     ao.bindEmitter(this)
     return fn.apply(this, arguments)
```
prettycoder: This looks magical ;) Where is `arguments` coming from?

prettycoder:
lgtm ;)

lib/probes/tedious.js
bmacnaughton: arguments is magical. it's the array-like object that is built in to JavaScript and contains all the arguments that were passed to a function (except arrow functions in which he still refers to the enclosing function's arguments).

so the arrow function `fn => function () {...}` is a function that accepts the argument `fn` and returns the function `function () {...}`. that function shows no arguments because it doesn't care (and because they might change). the only purpose of wrapping `f` is to bind `this` within the constructor. so `function() {...}` just binds `this` within `connection`'s constructor when it is calling `f`. then it goes ahead and calls `fn` normally. there are two forms of function calls in javascript when one wants to pass arbitrary arguments `fn.call()` and `fn.apply`. the first argument is the `this` to use. `call` enumerates each argument individually while `apply` has one argument which is an array (or array-like object) with all the arguments.

i'm not sure that makes it more clear or not. this was odd to me because it wasn't really caring about patching the function `fn`. it's only doing that so it can bindEmitter on `this`.

prettycoder: Thanks, helps with understanding for sure. 
It's like with some foreign languages, I'm starting to understand but couldn't speak it.

PR #38, opened 2018-11-14
Fix mongodb-core tests to use unique DB names [AO-8024]

bmacnaughton:
When running multiple matrix tests in parallel the mongo tests would cause others to fail because they all used the same database and collection names.

mongodb-core is the last of the database tests to use unique names. All the tests to depend on the environment variable `AO_IX` to be set - that is used to construct the unique names. Running locally there is no need to set `AO_IX` because only one test runs at a time. These commits just add some debugging output that allows makes it easier to see what is happening.

### DIFF
--- a/test/probes/mongodb-core.test.js
+++ b/test/probes/mongodb-core.test.js
@@ -130,7 +130,6 @@ function makeTests (db_host, host, isReplicaSet) {
     })
 
     ao.logLevel = 'error,warn,debug,patching'
-    ao.loggers.debug(`using AO_IX=${process.env.AO_IX}`)
 
     let server
     if (hosts.length > 1) {
--- a/test/probes/mongodb-core.test.js
+++ b/test/probes/mongodb-core.test.js
@@ -129,6 +129,9 @@ function makeTests (db_host, host, isReplicaSet) {
       }
     })
 
+    ao.logLevel = 'error,warn,debug,patching'
+    ao.loggers.debug(`using AO_IX=${process.env.AO_IX}`)
+
     let server
     if (hosts.length > 1) {
       const options = {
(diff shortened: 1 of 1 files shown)


test/probes/mongodb-core.test.js
```diff
@@ -129,6 +129,8 @@ function makeTests (db_host, host, isReplicaSet) {
       }
     })
 
+    ao.logLevel = 'error,warn,debug,patching'
```
prettycoder: in other tests you save the previous loglevel and reset it later. Needed or not needed here?
why this selection of levels, when only 'debug' is used in this file? Is there logging happening somewhere else. If so, why is 'info' excluded?

test/probes/mongodb-core.test.js
```diff
@@ -245,7 +254,16 @@ function makeTests (db_host, host, isReplicaSet) {
         steps.push(exit)
 
         helper.test(emitter, function (done) {
-          db.command(`${dbn}.$cmd`, {create: `coll-${dbn}`}, done)
+          db.command(`${dbn}.$cmd`, {create: `coll-${dbn}`},
+            function (e, data) {
```
prettycoder: right now only the collections tests have error logging. Is that the intention? If so, why is it needed here?

bmacnaughton:
The environment var isn't necessary - I could use pid. I might change that in the future as it's a bit more bulletproof for situations where only a single container is restarted. I don't know why I didn't think of using pid - seems obvious right now.

test/probes/mongodb-core.test.js
bmacnaughton: `error,warn` is the default and I want to keep them active. `patching` is used when a probe is loaded by the javascript `require` function (more precisely, our replacement of the native `require` function). `info` is not in the default set so it's not included.

logging overall needs to be scrubbed but i couldn't really have done it before i understood the use cases.

test/probes/mongodb-core.test.js
bmacnaughton: the only reason is that the collections test were where i was seeing errors to begin with so i wanted to get more information about them. because they're tests i figured there was no harm leaving them in - there is no production cost and getting additional information when tests fail is usually a pretty good thing.

PR #41, opened 2018-11-28
https fixes & new testeachversion

bmacnaughton:
- fix node 10.9.0 changed the http/https `request()` signature. handle the new signature.
- fix node 9.9.0+ `https.get()` no longer calls `exports.request()` so we failed to trace `https.get()` calls.
- add tests for the different `request()` signatures.
- fix logging error when using `%l` with a null/undefined layer.
- fix error in test code where test ended before the response could be received.
- update `versions.js` file for new `testeachversion`
- change `invalid header` logging message to not require formatting by `debug`. change test for it.
- update `devDependencies` to latest version supported when using node 8.


lib/probes/http.js
```diff
@@ -35,14 +34,37 @@ function patchClient (module, protocol) {
...
+
+    // the first argument might be a string or an url.URL
+    if (typeof args[0] === 'string') {
+      const urlString = args.shift()
+      options = url.parse(urlString)
+    } else if (args[0] instanceof url.URL || args[0].searchParams) {
+      options = urlToOptions(args.shift())
+    }
+
+    if (args[0] && typeof args[0] !== 'function') {
```
prettycoder: `else if` ?

lib/probes/http.js
```diff
@@ -53,34 +75,40 @@ function patchClient (module, protocol) {
...
 
       // Parse options object
-      if (typeof options === 'string') {
-        options = url.parse(options)
+      /*
+      if (typeof urlstring === 'string') {
+        urlopts = url.parse(urlstring)
       }
       const parsed = extend({}, options)
+      // */
```
prettycoder: this commented out code can probably go because it is on line 49.. now
It seams a bit out of context here.

lib/probes/http.js
```diff
@@ -53,34 +75,40 @@ function patchClient (module, protocol) {
...
 
       // Add X-Trace header to trace hops
       options.headers = options.headers || {}
       options.headers['X-Trace'] = span.events.entry.toString()
 
       // Set default protocol
-      parsed.protocol = parsed.protocol || protocol + ':'
+      options.protocol = options.protocol || protocol + ':'
 
+      let filtered = options
```
prettycoder: as I understand `let` it creates a reference, so 'filtered' references 'options'
on line 102 'filtered' may become a copy of 'options'
and then both are used to extract entry event info.
I don't quite understand why a reference and sometimes a copy is needed, since 'options' is scoped to this block. Maybe because of refactoring it got like this.

lib/probes/http.js
```diff
@@ -149,9 +179,13 @@ function patchClient (module, protocol) {
   // wrap request for http and https
   shimmer.wrap(module, 'request', wrapper)
 
-  // only wrap get for http if at least node 8
-  if (protocol === 'http' && semver.satisfies(process.version, '>=8')) {
-    shimmer.wrap(module, 'get', wrapper)
+  // wrap get if at least node 8. there are a few versions
```
prettycoder: This comment is a bit hard to understand. I don't see https for Node 8 being wrapped.
I would add what it is in the PR comment "Node 9.9.0+ https.get() no longer calls exports.request()" (... so we need to  wrap get for http and https )

bmacnaughton: funny, i was looking at this comment and thinking the same thing. i didn't update it when i made the code changes. the issue is that back in 8.0.0 (maybe 7.x but we don't support non-LTS versions) node changed the way `get()` worked for http but not for https. then they changed https in 9.9.0. my original fix was just to change it for both based on node > 8.0.0 but that broke things. the comment is leftover from the first abortive fix.

thanks!

PR #45, opened 2018-12-13
Docker test updates

prettycoder:
[testing] a few tweaks to get a docker test environment on Mac OSX

- env vars that refer to other containers are set in docker-compose
- APPOPTICS_SERVICE_KEY is set on the host, to avoid confusion when
  multiple people are sending traces
- upgraded to node 10

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,50 +3,66 @@
 #
 version: '2.1'
 
+# run with:
+# > docker-compose run --service-ports --rm --name node_main node_main /bin/bash
+
 services:
-#  main:
-#    container_name: main
-#    image: main
-#    build:
-#      context: test/docker
-#      dockerfile: main.yml
-#      args:
-#        # if AO_TEST_PACKAGE and AO_TEST_GITAUTH are not set then the
-#        # bindings installation will attempt to download the HEAD of the
-#        # master branch from the public repository.
-#        #
-#        # expect librato/node-appoptics-bindings#new-liboboe while private
-#        - AO_TEST_PACKAGE
-#        # expect a git auth token (or extend Dockerfile with user and password)
-#        - AO_TEST_GITAUTH
-#        # - AO_TEST_GITUSER
-#        # - AO_TEST_GITPASS
-#        # - AO_TEST_COLLECTOR_CERT=${AO_COLLECTOR_CERT:-test/certs/scribe-collector.crt}
-#        - AO_TEST_COLLECTOR_CERT=${AO_COLLECTOR_CERT:-test/certs/java-collector.crt}
-#        # - AO_TEST_COLLECTOR=${AO_COLLECTOR:-scribe-collector:4444}
-#        - AO_TEST_COLLECTOR=${AO_COLLECTOR:-java-collector:12222}
-#        # the next two set liboboe to send to the mock collector
-#        # for testing. these should be set to run the test suite.
-#        - AO_TEST_REPORTER=udp
-#        - AO_TEST_REPORTER_UDP=localhost:7832
-#    volumes:
-#      # map the directory this file is in to the /appoptics/ directory
-#      - "${PWD}:/appoptics/"
-##    volumes:
-##      - type: volume
-#    logging:
-#      options:
-#        max-file: "1"
-#        max-size: 50m
+  node_main:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -84,6 +100,7 @@ services:
   # the correct image.
   #
   cassandra:
+    container_name: "node_cassandra"
```
bmacnaughton: not sure you need to do this do you? when i use docker-compose up -d all the containers get the prefix of the directory i am in, e.g., ao8 or ao10. i'm thinking that you will use a different directory for ruby vs. node vs. whatever, so they would all be named differently regardless, right?

docker-compose.yml
```diff
@@ -3,50 +3,67 @@
 #
 version: '2.1'
 
+# run with:
+# > docker-compose run --service-ports --rm --name node_main node_main /bin/bash
```
bmacnaughton: Is there a way to run this without starting main or waiting on it?

what does `--rm` do?

also, i've never used `docker-compose run` just `docker-compose up -d` or `docker-compose down`. is it pretty much the same as `docker run -it node_main /bin/bash`?

test/docker/main.yml
```diff
@@ -25,7 +25,7 @@ RUN apt-get update && apt-get -y install gcc-4.9 g++-4.9 \
 && rm -rf /var/lib/apt/lists/*
 
 # get node
-RUN curl -sL https://deb.nodesource.com/setup_4.x | bash -
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -
```
bmacnaughton: version 4 is definitely outdated so it should be updated. i generally try to develop in the version prior to the latest LTS version mostly so it's less likely that any unsupported features will creep into the code and only be discovered during testing potentially forcing rethinking the implementation.

i'm not going to be using this development environment, so this if you like running at the most recent LTS version that works for me.

prettycoder: I was under the impression that 11 is the latest version, but it is 'current release' not LTS.
I'll change to it to 8.

docker-compose.yml
prettycoder: I was under the impression that you don't really use docker containers. I like them named, it is more predictable. Do you need the folder name prepended, because you test with different versions?
My folder name is 'appoptics-apm-node', they also get a count appended. So it is `appoptics-apm-node_memcached_1` versus  `node_memcached` when named.
Not sure about -p, it is for the port when starting up with `run`.
Why should the mysql container name be removed?

docker-compose.yml
prettycoder: `docker-compose up -d` works too, it starts the main container with the command defined in the Dockerfile. 
I like to have this command handy because it gets me into a shell.
It is like `run -it` but starts up the container, the network, and all the container it depends on. When you ctrl-d it doesn't shut everything down, but --rm removes this container and this command can be used to start it and join the network again. It may work without --rm, but it is not documented that way.
I like this way of starting up because instead of /bin/bash it can also be a script. I'm working on running `. env.sh docker` on startup. In ruby I have a script that will set up everything and depending in the argument will run all tests or start a shell.

docker-compose.yml
bmacnaughton: i use docker containers to run all the databases, i.e., everything other than my development environment. it's easy enough to change the name of the directory to `node` if you want that prefix. i just don't want it. and yes, it has a `_1` but that's never bothered me either.

docker-compose -p is different than docker -p: https://docs.docker.com/compose/reference/overview/

i'm not sure the interplay with `docker-compose run` and `docker-compose up -d` (the only command i use). it seems like `docker-compose up -d` will start a main container, so you'd be `docker exec -it ao8_main_1 /bin/bash` (or something like that) to connect to the already running container. or maybe you use `docker-compose run ...` without starting the other containers?

maybe it makes sense for you to add `-p node` if you want to change the naming? while i rarely have any conflicts i don't see the benefit in hard coding the name. you can use `docker-compose -p node run -p x:y /bin/bash` or something like that.

bmacnaughton: Oh, and I'd remove the mysql container name just so it's consistent with the others.

prettycoder:
not doing it this way

PR #50, opened 2019-01-11
Mongo updates

bmacnaughton:
- handle old `Group_Reduce` KV source in addition to new.
- add `aggregate` `QueryOp` (stop synthesizing `count`)
- handle class and object versions of `bindings.Reporter` (allowed testing with N-API bindings)
- minor binding error message changes

### DIFF
--- a/lib/probes/mongodb-core.js
+++ b/lib/probes/mongodb-core.js
@@ -255,6 +255,8 @@ const dataMakers = [
       data.Group_Reduce = cmd.group.$reduce.toString()
     } else if (typeof cmd.group.$reduce === 'object') {
       data.Group_Reduce = cmd.group.$reduce.code.toString()
+    } else {
+      data.Group_Reduce = cmd.group.$reduce.toString()
     }
     data.Group_Key = JSON.stringify(cmd.group.key)
   }],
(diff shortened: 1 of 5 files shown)


lib/probes/mongodb-core.js
```diff
@@ -255,6 +255,8 @@ const dataMakers = [
       data.Group_Reduce = cmd.group.$reduce.toString()
     } else if (typeof cmd.group.$reduce === 'object') {
       data.Group_Reduce = cmd.group.$reduce.code.toString()
+    } else {
```
prettycoder:
```
 if (typeof cmd.group.$reduce === 'function') {
      data.Group_Reduce = cmd.group.$reduce.toString()	    
    } 
```
this is now also captured in the else clause, it seems that the test for 'object' is sufficient to cover the 2 cases.

lib/probes/mongodb-core.js
```diff
@@ -267,14 +269,8 @@ const dataMakers = [
...
-    const p = cmd.pipeline
-    if (p[0].$match && p[1].$group) {
-      const p1 = p[1].$group
-      if (p1._id === null && p1.n && p1.n.$sum === 1) {
-        data.QueryOp = 'count'
-        data.Query = JSON.stringify(p[0].$match)
-      }
-    }
+    data.QueryOp = 'aggregate'
+    data.Pipeline = JSON.stringify(cmd.pipeline)
```
prettycoder: nice! I double-checked with the spec ;)

test/probes/mongodb.test.js
```diff
@@ -154,7 +154,30 @@ function makeTests (db_host, host, isReplicaSet) {
...
+      server = new mongodb.ReplSet(servers, options)
+      const mongoClient = new MongoClient(server, {})
+
+      mongoClient.connect((err, _client) => {
+        ao.loggers.test.debug('mongoClient() connect callback', err)
+        if (err) {
+          console.log('error connecting', err)
+          return done(err)
+        }
+        ctx.server = server
```
prettycoder: where is `ctx` defined?

lib/probes/mongodb-core.js
bmacnaughton: I updated the spec to conform to this, so it's a little bit self-referential.

test/probes/mongodb.test.js
bmacnaughton: line 90.

PR #51, opened 2019-01-15
Add support for koa-router >=7.0.0

chooban:
In order to evaluate the AppOptics service (still ongoing) I have to update the node library to work with koa-router >= 7.0.0. This is what I've got so far so please consider this WIP, but I'm happy to make changes to fit in with better practices, or simply hand it over if you'd prefer to implement the changes differently. 

The most WIP aspect is definitely the testing as it fails. I get this, which I'm not sure about:

```
Uncaught AssertionError: expected Object {
  Edge: '96F5D47AEEC559FC',
  Hostname: 'a6bb577c405f',
  Label: 'exit',
  Layer: 'koa',
  TID: 40,
  Timestamp_u: 1547563758829518,
  'X-Trace': '2B366D0D686BD1363D43CF16E7702B281EEEA953383EE5E6EC5D598D9901',
  _V: '1'
} to have property Label of 'profile_entry' (got 'exit')
```
I imagine it's something fairly obvious to those in the know. The updates appear to be working in our app so it seems to work.

The version number got bumped by me just so that I could double check I had my version installed after forking. 

I had to install a later version of node as the docker image building failed otherwise.

Thank you for the Mac setup, it was very helpful!

### DIFF
--- a/lib/probes/koa-router.js
+++ b/lib/probes/koa-router.js
@@ -2,13 +2,27 @@
 
 const shimmer = require('ximmer')
 const methods = require('methods')
+const semver = require('semver')
+
 const utility = require('../utility')
 const ao = require('../')
+const requirePatch = require('../require-patch')
 const Span = ao.Span
 const conf = ao.probes['koa-router']
 
-module.exports = function (router) {
-  return app => patchApp(app, router)
+const {version} = requirePatch.relativeRequire('koa-router/package.json')
+
+module.exports = function (_router) {
+  if (semver.gte(version, '7.0.0')) {
+    // allow the caller to invoke this with or without "new" by
+    // not using an arrow function.
+    return function () {
+      const router = _router.apply(this, arguments)
+      wrapRouter(router)
+      return router
+    }
+  }
+  return app => patchApp(app, _router)
 }
 
 function patchApp (app, router) {
@@ -17,7 +31,7 @@ function patchApp (app, router) {
   methods.concat('del').forEach(method => {
     // For 5.x+, the methods are on a router, not patched onto the app
     //Check if app exists, only use handler otherwise
-    patchMethod(
+    patchGeneratorFunction(
       (app ? [handler, app] : [handler]).filter(v => typeof v[method] === 'function')[0],
       method
     )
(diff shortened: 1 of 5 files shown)


bmacnaughton:
Thanks for the PR. I'm out right now but will try to take a look tomorrow.

chooban:
Sounds good. I also have a minimal test repo for a koa server: https://github.com/chooban/koa-router-appoptics

bmacnaughton:
I really appreciate your contributions to this. I am not at all an expert in koa and it's clear that you're proficient with it. I have yet to look into the tests but I only see one issue - we can't use the keywords `async` or `await`.

I know that's kind of a big issue. I've wrestled with it in a places. First, in testing, I used `eval()` to define the function to be used when it was node 8 or higher; lower versions wrapped the callbacks with a promise-returning function. That's pretty ugly. After that I just split it into separate files (see v5.9.3's postgres testing). The level of complexity that comes from supporting events, callbacks, generators, and promises made the choice easy for me.

Feel free to make changes so the code uses promises as opposed to async/await, or, if you'd prefer, I'll work on it after I wrap up what I'm working on.

I'm interested in some background - it looks like you're using promises (async/await). I understand why a number of packages went to generators before async/await and the standardization of promises. Do you know whether a sizable portion of people are using generators? I'm asking because I've seen multiple packages drop support for them.

chooban:
I'll see if I can update it to not use async/await. Is that because of the versions of node that need to be supported?

I'm not aware of any major libraries or frameworks that still use generators as their preferred method if they can also move to async/await. I think people picked up on that much more easily than generators.

bmacnaughton:
> Is that because of the versions of node that need to be supported?

Yes, version 6 of node doesn't parse async/await and generates syntax errors.

bmacnaughton:
> The most WIP aspect is definitely the testing as it fails. I get this, which I'm not sure about:

On the error, I'll take a look when I get a chance but the it looks like it's failing to create a profile span (a special kind of span) when it should be. When you get an "exit" instead of an "entry" it generally means that a span wasn't created for some reason.

The test logic intercepts messages sent by the agent (through a non-straightforward path) and checks them against expected messages (not all properties, but key ones).

I find it easier to test and debug using mocha directly (i.e., not through gulp). I haven't changed docs yet but just running `mocha test/koa-router.test.js` should do the trick. If you haven't put `node_modules/.bin` in your path you either need to do that or prefix mocha as `node_modules/.bin/mocha`. This lets you `mocha --inspect-brk test/koa-router.test.js`. Some test modules might need to include `const should = require('should')`. I've been changing them as I work with them to use `const expect = require('chai').expect` because it works even when the value is undefined and doesn't require special casing.

Sorry if some of this is old hat - I just want to make sure you're not fighting too many things at once.

chooban:
Should I be downgrading the node version to 4 or 6? I reverted my change so that the `main.yml` installs version 4 but I then get problems as a lot of the imports use object destructuring which only came in with version 6.

bmacnaughton:
Just downgrade to 6; we no longer support node version 4.

bmacnaughton:
fyi, i'm working on this now as we have another customer waiting on it as well. i'm going to ask you to review my changes and, if you don't mind, will put you in the contributors list - your knowledge has been a good jump start for me.

if you're interested at some time i'd be happy to do a screen share and show you how i go about debugging this kind of thing.

chooban:
All sounds good to me! I think our trial period expires this evening, so I'd have found checking my changes to be tricky. 

I'm in the UK, but I'd be interested to see how you go about it at some point.

cheempz:
Hi @chooban, can you let me know your AppOptics organization name or the email you use to log in?  I'd like to extend the trial for you :)

cheempz:
FYI @chooban if you prefer to send this info over email, please find me at lin@solarwinds.cloud

bmacnaughton:
Hey, if you get any time take a look at https://github.com/appoptics/appoptics-apm-node/pull/52 and let me know if I've missed anything obvious. I appreciate your help.

chooban:
Can do. I should have some time tomorrow to try this version of the library with our app and make sure it all works as expected.

PR #53, opened 2019-02-08
v6.0.0 - sample check inversion, profile => spans, major overhaul, and cleanup

bmacnaughton:
1. removed RUM artifacts
2. removed Profiles
3. moved sampling decision from Event constructor to function calling Span constructor.
4. [API BREAKING] changed API signature of Span
   - 3rd argument is settings object, not xtrace/parent
   - 3rd argument must contain {inbound, doSample}
   - 3rd argument optional {metadata}
   - inbound (boolean) the span is a top level span as the result of an inbound request. it will either start a trace or continue one from an external source.
   - doSample (boolean) sample this trace or not.
5. [API BREAKING] changed API signature of Event
   - 3rd argument is metadata to use for the event creation.
   - 4th argument is whether to edge back to the metadata or not. (might be able to avoid this).
6. stringToMetadata() consider error if opID is all zeros (call lower level bindings fn to avoid)
   - takes check out of probes/http.js
7. packages with probe naming considerations (and their tests)
    - director
    - express
    - koa-resource-router
    - koa-route
    - koa-router
    - restify
8. moved KV pairs out of event and into event.kv https://chromium.googlesource.com/external/github.com/v8/v8.wiki/+/60dc23b22b18adc6a8902bd9693e386a3748040a/Design-Elements.md
9. stubbed new function bindings.Context.getTraceSetting() for use by entry/root spans.
10. added XgetTraceSettings() to appoptics-bindings/napi-branch. will become replacement for all entry/root spans.
11. [API BREAKING] change signature of span builder function in startOrContinueTrace, instrument, runInstrument, instrumentHttp. This impacts many tests.
12. move all decision/metadata generation logic into bindings.Context.getTraceSettings(). (this is implemented in index.js until there is a production tested oboe ready to go with the API in it. when that's ready the code will start using the oboe facility and the javascript implementation can be removed.)
13. modified span builder function return value's finalize function: spanInfo.finalize(span, previousSpan)

(this is not ready to merge yet - depends on appoptics-bindings v6 for correct metrics decisions. in process of running the test matrices for this massive change. first release will be 6.0.0-rc and tagged "rc" so as not to be default install.)


bmacnaughton:
I created the 6.0.0 branch to perform a selective squash merge (`git rebase -i master`) with so this branch is no longer needed.)

PR #68, opened 2019-05-30
v6.5.0

bmacnaughton:
- implement environment variable handling
- log previously loaded packages as warn (was error)


prettycoder:
looks good to me, just wondering if some tests should be added.
this is what I did for Ruby, mainly check defaults and service_key. 
https://github.com/appoptics/appoptics-apm-ruby/blob/master/test/unit/oboe_init_options_test.rb

maia:
@bmacnaughton I assume you've noticed that Maia has a different handle here on github. All the best!

bmacnaughton:
this has been supplanted by the 6.5.1 release.

PR #73, opened 2019-07-27
use native function for prepending listeners [AO-12085]

bmacnaughton:
- use emitter.prependListener()
- replaces event-pre-handler.js/patchEmitter()

I'm going to base trigger trace work on this branch because they both fiddle primarily with lib/probes/http.js.

### DIFF
--- a/lib/probes/http.js
+++ b/lib/probes/http.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const patchEmitter = require('event-pre-handler')
 const shimmer = require('ximmer')
 const url = require('url')
 const os = require('os')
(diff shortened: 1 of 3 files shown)


lib/probes/http.js
```diff
@@ -344,12 +333,13 @@ function patchServer (module, protocol) {
 
   function wrapRequestResponse (span, req, res) {
     // Report socket errors
-    req._preHandle('error', error => span.error(error))
-    res._preHandle('error', error => span.error(error))
+    req.prependListener('error', error => span.error(error));
```
prettycoder: Why is this line repeated?

bmacnaughton: one is on the `req` object and the other is on the `res` object.

prettycoder: dahhh, I looked at it so many times ...

PR #78, opened 2019-10-02
update both cfg and config with serviceKey

bmacnaughton:
- ao exposes `cfg` but only `config` was being updated. both are updated now.
- clarify debug message


bmacnaughton:
This is incorporated into a broader PR, #79

PR #79, opened 2019-10-09
Rationalize configs

bmacnaughton:
Move all configuration (definitions, config file, environment variables) to single module. [AO-13820]

https://swicloud.atlassian.net/browse/AO-13820

- delete legacy configuration modules
- add unified-config.test.js for new config

[AO-13820]: https://swicloud.atlassian.net/browse/AO-13820


prettycoder:
can you add the jira ticket id somewhere?

lib/probe-defaults.js
```diff
@@ -0,0 +1,234 @@
...
+    collectBacktraces: true,
+  },
+
+  //
+  // context maintaining modules. they cannot be disabled and are listed
+  // for completeness.
+  //
+  'aws-sdk': {enabled: true},
+  bcrypt: {enabled: true},
+  bluebird: {enabled: true},
```
prettycoder: do the bluebird, generic-pool and q instrumentations need a check for enabled as well?

bmacnaughton: no, they don't. but you spotted an inconsistency that i decided to skip over - aws-sdk is not really in the same category as the others but i didn't want to create a separate category for it.

aws-sdk was created to fix apex learning's problem. when there was an error on the http rest call to aws the aws code would resend the request but would not recalculate the signature. but the probe didn't know it was a resend and so attached a different x-trace header to the request causing it to fail the integrity check at the aws end. so, yes, that probe can be disabled but 1) it's not really a probe, it's just adds 'x-trace' to the headers that aws-sdk ignores and 2) even though it shouldn't be disabled i thought it a good idea to allow a quick disable in case there was a problem with aws-sdk at some point in the future - it would be a quick diagnostic.

the others simply maintain context which needs to be done; disabling them would cause many traces to break because they'd just lose content. again, they're not really probes. it could be argued that generic-pool should become a probe as the queue time might be interesting. but i'll cross that bridge later.

test/unified-config.test.js
```diff
@@ -0,0 +1,453 @@
...
+  //
+  // config file errors
+  //
+  it('should correctly report an error reading a non-default config file', function () {
+    const file = process.env.APPOPTICS_APM_CONFIG_NODE = 'xyzzy-bruce.json';
+    const fullpath = `${process.cwd()}/${file}`;
+
+    const cfg = guc();
+
+    const message = 'Cannot find module \'/home/bruce/solarwinds/ao-agent/xyzzy-bruce.json\'';
```
prettycoder: Should this be generalized for other filesystems/users? `${process.cwd()}/${file}`

test/unified-config.test.js
```diff
@@ -0,0 +1,453 @@
...
+    doChecks(cfg, {file: fullpath, errors});
+  })
+
+  it('should correctly report an error reading a bad config file', function () {
+    writeConfigLiteral('{"i am: bad\n');
+
+    const cfg = guc();
+
+    const errors = [[
+      'Cannot read config file /home/bruce/solarwinds/ao-agent/appoptics-apm: ',
```
prettycoder: there is a duplication here and I think the filename should have the extension '.js'
Same issue with the path as previous test.

test/unified-config.test.js
```diff
@@ -0,0 +1,453 @@
...
+  it('should warn about unknown APPOPTICS_ environment variables', function () {
+    process.env.APPOPTICS_YINYIN_ROCKS = 'truth';
+
+    const cfg = guc();
+
+    const unusedEnvVars = ['APPOPTICS_YINYIN_ROCKS=truth'];
+    doChecks(cfg, {unusedEnvVars});
+  })
+
+  it('should not set use environment variable values', function () {
```
prettycoder: 'set use' ? It should not 'set invalid' env var values .... ?

test/unified-config.test.js
```diff
@@ -0,0 +1,453 @@
...
+    const config = {triggerTraceEnabled: false};
+    const warnings = ['APPOPTICS_TRIGGER_TRACE is deprecated; it will be invalid in the future'];
+    doChecks(cfg, {global: config, warnings});
+  })
+
+  //
+  // probes
+  //
+  it('should set probe values correctly', function () {
+    const config = {probes: {fs: {enabled: false, bruce: 'says hello'}}};
```
prettycoder: why does it not warn about invalid config name?

test/unified-config.test.js
```diff
@@ -0,0 +1,453 @@
...
+    const config = {transactionSettings: {type: 'url', string: 'i\'m a shrimp', tracing: 'disabled'}};
+    writeConfigJSON(config);
+
+    const cfg = guc();
+
+    const transactionSettings = toInternalTransactionSettings(config.transactionSettings);
+    doChecks(cfg, {transactionSettings});
+  })
+
+})
```
prettycoder: very thorough!
The only additional test I could think of is when tracing is disabled and there is a setting to enable certain urls. This would enable dowample and dometrics.

test/unified-config.test.js
bmacnaughton: thanks.

test/unified-config.test.js
bmacnaughton: i don't follow - why should it warn about an invalid config name?

test/unified-config.test.js
bmacnaughton: i don't think that should be covered by this test. the designed goal of this module is to process the configuration file and env vars in a consistent and predictable priority. getting into the semantics of what the properties mean is beyond the scope. for example, the module doesn't verify that a the serviceKey is formatted correctly - that is handled by the caller (index.js).

make sense? or am i off base with my thinking?

test/unified-config.test.js
bmacnaughton: generalization makes sense - thanks. i already have constructed fullpath = `${process.cwd()}/${file}` so i'm just using that. now it will handle macs. i think windows accepts `/` even though they love to output `\` but i'll cross that bridge when i get to it.

test/unified-config.test.js
bmacnaughton: i don't follow the duplication. what do you mean?

the filename shouldn't have the extension `.js` because the message is echoing the default name used. that does not specify an extension (so it will find `.js` files before `.json` files). but the later portion of the message is the error that node throws with the specific file that it was reading.

test/unified-config.test.js
prettycoder: My thinking about tests is that they should discover if some refactoring goes wrong. So I would like to have a test that goes through the enabled case, because there is a tiny bit of code that is different from the disabled case.
```
if (setting.tracing === 'enabled') {
      goodEntry.doSample = goodEntry.doMetrics = true
    }
```

bmacnaughton: i added the coverage on line 370 (the "should allow JSON x-action settings" test). no real reason for a new test (i don't think). thanks!

PR #81, opened 2019-10-15
lib/probe/http - prevent undefined TransactionName

bmacnaughton:
- supply a default when there is no TransactionName. [AO-14454]
- prepare for a numeric code return on errors [AO-14437]
- debounce the error logging.

The second item requires an appropriate release of appoptics-bindings.

[AO-14454]: https://swicloud.atlassian.net/browse/AO-14454

[AO-14437]: https://swicloud.atlassian.net/browse/AO-14437

### DIFF
--- a/lib/probes/http.js
+++ b/lib/probes/http.js
@@ -11,6 +11,9 @@ const Event = ao.Event
 
 const log = ao.loggers
 
+// avoid issuing too many errors on bad transactions
+const dbSendError = new log.Debounce('error');
+
 const ule = Symbol('UnexpectedLastEvent');
 
 const defaultPort = {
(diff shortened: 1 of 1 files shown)


lib/probes/http.js
```diff
@@ -523,24 +526,26 @@ function patchServer (module, protocol) {
...
-          && txname.indexOf(args.txname) + args.txname.length !== txname.length) {
-          log.warn(
-            'sendHttpSpan() changed TransactionName from %s to %s',
-            args.txname,
-            txname
-          )
-        }
+        // TODO BAM the txname not null check can be removed once appoptics-bindings has been
+        // changed to return the integer error code instead of a null string.
+        if (typeof txname === 'string' && txname) {
```
prettycoder: Detail: Should the null check come first?

lib/probes/http.js
```diff
@@ -523,24 +526,26 @@ function patchServer (module, protocol) {
...
-        // what to do if an error sending? don't know transaction name but shouldn't
-        // matter, so don't use the return value.
-        if (txname) {
-          exitKeyValuePairs.TransactionName = txname
+          // if it's a string the worst it can be is an empty string.
+          exitKeyValuePairs.TransactionName = txname;
         } else {
-          log.error('sendHttpSpan() returned empty TransactionName')
+          exitKeyValuePairs.TransactionName = args.txname || 'unknown';
+          dbSendError.log(`sendHttpSpan() code ${txname}`);
```
prettycoder: out of curiosity: When are these Debounced logs sent?

bmacnaughton: the default debouncing window is 1 second or 100 instances of the error. they're settable at construction time. right now i should fix the Debounce code to reset both `lastTime` and `lastCount` when a log is emitted; currently they're independent.

PR #90, opened 2019-11-13
zlib - fixes [AO-11896, AO-11898]

bmacnaughton:
- add Brotli support
- fix "event already sent error in zlib test"
- test Brotli compression

### DIFF
--- a/test/probes/zlib.test.js
+++ b/test/probes/zlib.test.js
@@ -3,9 +3,14 @@
 const helper = require('../helper')
 const {ao} = require('../1.test-common.js')
 const noop = helper.noop
+const once = require('../../lib/utility').once;
 
 const concat = require('concat-stream')
 const zlib = require('zlib')
+
+const expect = require('chai').expect;
+
+const semver = require('semver');
 
 const classes = [
   'Deflate',
@@ -26,6 +31,13 @@ const methods = [
   'inflateRaw',
   'unzip'
 ]
+
+if (semver.gte(process.version, '11.7.0')) {
+  classes.push('BrotliCompress');
+  classes.push('BrotliDecompress');
+  methods.push('brotliCompress');
+  methods.push('brotliDecompress');
+}
 
 describe('probes.zlib once', function () {
   let emitter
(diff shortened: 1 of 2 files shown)


test/probes/zlib.test.js
```diff
@@ -138,128 +150,157 @@ describe('probes.zlib', function () {
...
-              checks.exit(msg)
-            }
-          ], done)
-        })
-      }
+
+      it(`should support ${method}`, function (done) {
+        // make sure the function exists
+        expect(zlib).property(method);
+        expect(typeof zlib[method]).equal('function');
```
prettycoder: Not sure this should be an assertion in a test. 
I think the instrumentation implementation should check if the function exists and tests should test the behavior if the function does not exist.

test/probes/zlib.test.js
```diff
@@ -138,128 +150,157 @@ describe('probes.zlib', function () {
...
-            function (msg) {
-              checks.exit(msg)
-            }
-          ], done)
-        })
-      }
+      it(`should support ${syncMethod}`, function (done) {
+        // make sure the function exists
+        expect(zlib).property(syncMethod);
+        expect(typeof zlib[syncMethod]).equal('function');
```
prettycoder: dito ...

test/probes/zlib.test.js
bmacnaughton: it originally just didn't do the test if the function was missing; that was clearly wrong as missing functions wouldn't be noticed at all. so rather than not testing i changed it to fail if the method was missing.

the instrumentation does check to verify that the function exists.

so the question is whether the test verifies that the function is present and then tests it or whether it just tests it. something is wrong if standard methods are not present. it seems like that's higher order than trying to run the test if they're missing; it's a more specific diagnostic of the problem.

what do you think?

prettycoder: yes, I thinks it is fine, I now understand the practicality of it. 
I'm still wondering what the behavior of the instrumentation is, when the function does not exist.

bmacnaughton: the function would just be ignored if it didn't exist.

see the wrap function (lines 83, 111) where a check is made to see if the method exists. if it doesn't exist then nothing is patched (because there is nothing to patch). a logMissing() entry is made (that's 'patching' level with the prefix of the module name, 'zlib' in this case.

PR #92, opened 2019-11-13
package.json - change license form [AO-14757]

bmacnaughton:
- avoid warnings about license due to form


bmacnaughton:
you can reference the license as a file too - that would be shorter than the URL.

PR #99, opened 2019-11-30
restify fixes and extensive context work

bmacnaughton:
This started out to fix restify but it turned out that involved working with context.

- supports restify >= 7.0.0
- suppress logging missing span during node bootstrap phase
- add createContext({newContext: true|false}) option
- add support for ace-context, remove continuation-local-storage
- add .mocharc with exit: true to prevent tests hanging
- ignore *.log files in package directory


lib/probes/restify.js
```diff
@@ -92,26 +85,55 @@ function patchServer (server) {
...
-function patchCreateServer (restify) {
+function patchCreateServer (restify, version) {
   if (typeof restify.createServer !== 'function') {
+    logMissing('restify.createServer()');
     return
   }
   shimmer.wrap(restify, 'createServer', fn => function () {
     const server = fn.apply(this, arguments)
-    patchServer(server)
+    patchServer(server, version)
```
prettycoder: doesn't look like the patchServer function takes a version arg

PR #101, opened 2019-12-12
add option to force a new trace to begin

bmacnaughton:
`startOrContinueTrace()`'s config option can now contain `forceNewTrace: true` in order to ignore any existing trace and unconditionally start a new trace.


bmacnaughton:
The jt/partial-websocket-support branch includes these changes. This branch does not need to be merged.

PR #104, opened 2019-12-20
tedious - handle no database specified

bmacnaughton:
- lib/probes/tedious - don't add KV if no database
- test/probes/tedious - test no database case

### DIFF
--- a/test/probes/tedious.test.js
+++ b/test/probes/tedious.test.js
@@ -12,6 +12,7 @@ if (!process.env.AO_TEST_SQLSERVER_EX) {
 
 const helper = require('../helper')
 const {ao} = require('../1.test-common')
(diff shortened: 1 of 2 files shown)


test/probes/tedious.test.js
```diff
@@ -184,17 +195,20 @@ describe('probes.tedious ' + pkg.version, function () {
...
       port: addr.port,
       options: {
-        database: 'test',
         tdsVersion: '7_1',
         encrypt: false
       }
-    })
+    };
+    if (dbname) {
+      options.options.database = dbname;
```
prettycoder: maybe `settings` (instead of 'options') for the const name would make this easier to read ?

bmacnaughton: makes sense.

PR #109, opened 2020-01-06
todo-cleanups [AO-15036]

bmacnaughton:
This PR is a melange of changes that are only related by the fact that they had a TODO comment. The goal of this release is to do one of:
- take the action specified by the TODO
- remove the TODO because it's no longer relevant
- enter a JIRA for the TODO because it's still relevant but not a priority to address

There are also a number of comments updated and minor edits for issues I ran into while looking into each TODO.


lib/probes/http.js
```diff
@@ -570,13 +570,13 @@ function patchServer (module, options, protocol) {
...
+        // error then it will be the txname that was used, which might not match
+        // the txname the user specified if it wasn't valid.
         const txname = ao.reporter.sendHttpSpan(args);
 
-        // TODO BAM the txname not null check can be removed once appoptics-bindings has been
-        // changed to return the integer error code instead of a null string (version > 7.1.0)
-        if (typeof txname === 'string' && txname) {
-          // if there is a txname and it doesn't match
+        if (typeof txname === 'string') {
+          // see if the name used doesn't match the name provided
```
prettycoder: on line 584 should it be `txname.endsWith(args.txname)`

bmacnaughton: yes, it should. thanks.

i think even more directly i can get rid of the domainPrefix check because the .endsWith will ignore that.
so: `if (!txname.endsWith(args.txname)) {` seems to work.

PR #116, opened 2020-04-10
add notifications, 

bmacnaughton:
- async notifications, including soft disable message
  - requires bindings v10 (currently alpha)
- add support for APPOPTICS_PROXY and proxy config file options
- remove span.descended in favor of !span.topSpan


lib/index.js
```diff
@@ -451,6 +458,25 @@ if (enabled) {
...
+    const fileOnlyPrefix = ao.notifications.socketPrefix;
+    files.filter(f => f.startsWith(fileOnlyPrefix))
+      .forEach(f => {
+        // if it's the currently active socket ignore it
+        if (ao.notifications.socket.endsWith(f)) {
+          return;
+        }
+        // delete the file ignoring any errors.
+        ao.fs.unlink(`${dir}${f}`, () => undefined);
+      });
```
prettycoder, line 477: Can there be multiple socket paths?
If so, should this be an exit from the loop instead of a return from the function?

lib/notifications-listener.js
```diff
@@ -0,0 +1,122 @@
...
+    }
+  } else if (msg.source === 'collector') {
+    //
+    // collector messages
+    //
+    if (msg.type === 'remote-config') {
+      // take appropriate action
+    } else if (msg.type === 'remote-warning') {
+      if (msg.message.startsWith('The API token supplied was not found.')) {
+        ao.loggers.error('Invalid serviceKey; AppOptics APM is disabled.');
```
prettycoder, line 97: this is decoupled from where the APM is disabled. How does node know to disable the APM?

test/solo/notifications-timeout.test.js
```diff
@@ -0,0 +1,50 @@
...
+      done();
+    }, 60000);
+  });
+
+  it('should handle restarting if keep-alive messages do not arrive', function (done) {
+    this.timeout(35000);
+
+    const timeToStop = Date.now() + 30 * 1000;
+    const iid = setInterval(function () {
+      // fake keep-alives not arriving my setting the last message time to
```
prettycoder, line 37: my -> by

test/solo/notifications-timeout.test.js
```diff
@@ -0,0 +1,50 @@
...
+        if (m.source === 'oboe' && m.type === 'keep-alive') {
+          keepAliveCount += 1;
+        }
+      });
+      expect(keepAliveCount).gte(4, 'at least 4 keep-alive messages');
+      done();
+    }, 60000);
+  });
+
+  it('should handle restarting if keep-alive messages do not arrive', function (done) {
```
prettycoder, line 32: how do we know it is restarting?

test/solo/notifications.test.js
```diff
@@ -0,0 +1,104 @@
+'use strict';
+
+// oboe's notifier will send all log messages
+process.env.APPOPTICS_DEBUG_LEVEL = 6;
+
+const ao = require('../..');
```
prettycoder, line 6: Is there some magic here? You can just require the root directory?

test/solo/notifications.test.js
```diff
@@ -0,0 +1,104 @@
...
+
+const messages = [];
+
+notifications.on('message', function (msg) {
+  messages.push(msg);
+})
+
+const expect = require('chai').expect;
+
+describe('notification function tests (long tests)', function () {
```
prettycoder, line 45: where are the messages coming from?

lib/index.js
bmacnaughton: it's possible that their might be more than one though unlikely in production. the cost of looping through them is low enough that i'll leave it in for now. sometimes when testing there is more than one. but you do make me wonder what might happen if multiple applications are running on one host - there could be more than one active socket path then; i'll need to test that the delete fails (because the file is in use) in that case.

lib/notifications-listener.js
bmacnaughton: it turns out that the agent doesn't need to do anything at all other than issue the message. oboe handles it by always returning no sample/no metrics when disabled.

test/solo/notifications-timeout.test.js
bmacnaughton: `(notifications.startCount > 1)`

test/solo/notifications-timeout.test.js
bmacnaughton: changed in current branch.

test/solo/notifications.test.js
bmacnaughton: maybe i've been missing something all along but how do i know what the root directory of the package is?

test/solo/notifications.test.js
bmacnaughton: lines 158, 162, or 172 of `lib/notifications.js` where it executes `this.emit()`.

events are essentially the same thing as callbacks except there can be multiple listeners on an event while a callback is a singleton.

lib/notifications-listener.js
prettycoder: I see :)

test/solo/notifications-timeout.test.js
prettycoder: Am I blind?

test/solo/notifications.test.js
prettycoder: I mean '../..' wouldn't that be the root/top directory of the repo?

test/solo/notifications.test.js
prettycoder: but where do the messages get created that are emitted? I think I'm missing something in the way this test is setup.

test/solo/notifications-timeout.test.js
bmacnaughton: it's a lot of new stuff to absorb. the mind overloads after a while.

test/solo/notifications.test.js
bmacnaughton: it is, but how else can i refer to the root/top of the repo?

test/solo/notifications.test.js
bmacnaughton: oh, i see. it counts on getting the messages from the production collector. one of the reasons these needed to be standalone tests is that they set the environment to use the production collector and also set oboe's log level to 6 in order to assure that the right messages are received. if things change then the checks for the test might need to change but that should be obvious then.

lib/index.js
bmacnaughton: interesting, but ok. i can delete an existing socket path but existing client-server connections continue to work. that's good. so if multiple applications are running on a single machine then the worst case is that a different application's socket path is deleted before the client attaches to it. in that case oboe will default logging. it's ugly but information isn't lost. if the path is deleted after the client-server connection is established then there is no problem. that's a pretty short window for an error; it could be mitigated by checking the creation time of the socket path but i suspect that it's unlikely that the problem will be hit. if that assumption is wrong the timestamp is a fix.

lib/notifications-listener.js
cheempz: Late to the party, I see that this has been merged but @bmacnaughton you need to change this behaviour -- the collector will send down "WARNING" log messages that specify why a service key is considered invalid.  This is critical to help support and us to diagnose the issue.

PR #117, opened 2020-04-21
general cleanup and span creation minimization

bmacnaughton:
- minimize span/event creation for unsampled traces
  - create entry span and a single "skeleton" span (like a skeleton key)
- env.sh - remove TOKEN_BUCKET env vars
- isReadyToSample() is directly on `aob` now, not `aob.Reporter`
- rework span.runAsync() and span.runSync()
  - default transaction names at span creation time.
  - simplify logic
- span.exitWithError() => span.exitCheckingError()
- getFormattedTraceId() uses lastEvent, not topSpan, for taskID and sample flag.
- cleanup event.js comments and cruft
- remove metadata formatters
- debounce metric send errors to avoid multiple reports each interval
- implement timeout recovery in notifications.js
- rework test.sh to run each test separately to avoid mocha-created contexts
- test updates for new bindings and changes listed above.
- new tests for skeleton spans.


test/composite/request.test.js
```diff
@@ -165,7 +159,7 @@ describe('probes.request', function () {
         res.end('done')
       })
 
-      const origin = new ao.Event('span-name', 'label-name', addon.Metadata.makeRandom(1))
+      const origin = new ao.Event('span-name', 'label-name', addon.Event.makeRandom(1))
```
prettycoder, line 162: there is one more like this in api-sim line 27

test/probes/restify.test.js
```diff
@@ -30,8 +30,9 @@ if (semver.satisfies(process.version, '>=8.0.0')) {
 
 describe(`probes.restify ${pkg.version}`, function () {
   let emitter
-  let fsState
+  let fsState;
```
prettycoder, line 33: well, the previous line could also end in `;` in this case

lib/api.js
```diff
@@ -151,13 +151,56 @@ function definePropertiesOn (ao) {
...
   Object.defineProperty(ao, 'lastEvent', {
-    get () {return Event && Event.last}
-  });
+    get () {
+      let last
+      try {
+        last = ao.requestStore.get('lastEvent');
+      } catch (e) {
+        // avoid logging during node bootstrap phase. https://nodejs.org/api/async_hooks.html.
+        if (asyncHooks.executionAsyncId() !== 1) {
```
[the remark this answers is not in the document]
bmacnaughton: sort of. it's more a matter of whether the async ID currently being executed is 1 which means that this is during the node bootstrap phase.  this is a big issue when debugging ace-context and the cost is a check only when there is a failure, and that minimizes false positive logging.

lib/notifications.js
```diff
@@ -176,15 +194,26 @@ class Notifications extends EventEmitter {
...
           this.server.close(e => {
             this.server = undefined;
             this.serverStatus = 'initial';
           });
-          throw e;
+          break;
         }
       }
     }
+    if (this.serverStatus === 'listening') {
```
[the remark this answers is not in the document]
bmacnaughton: when i wrote the comment i was throwing the exception. that seems like a bad idea because the fallback is just that oboe will log as usual and the agent shouldn't, in general, ever throw an exception. so i changed throwing the exception to a `break` and check the status at the end. if not listening then it closes the server and returns the server status. so now the comment really should be moved to the `else` on 289 and changed to wonder whether i should log it or not. i'll try to add that in a subsequent pr.

test/probes/restify.test.js
bmacnaughton: yes, it could have eagle-eye. i would really like to go through module by module and make that change with *no* other changes.

bmacnaughton:
on span.test.js - the section that begins with
```
  //========================================================================
  //========================================================================
  // skeletonized span handling. these are unsampled traces that create an
  // entry span and a single skeleton span that is used for all other spans.
  //========================================================================
  //========================================================================
```

would be useful to give a bit of a look. my goal there was to verify that different combinations of sync and async spans could all use the skeleton span without problem. the reason is that when `span.descend()` is called all it does is either 1) create a skeleton span and return it (if the current span is a topSpan) or 2) return the skeleton span itself (if not a topSpan). so making sure that doesn't create problems was the goal of the tests.

PR #118, opened 2020-04-22
deprecate Event.last and Span.last

bmacnaughton:
- Event.last => ao.lastEvent
- Span.last => ao.lastSpan

the base of this branch is `p9p-noti+minimal`; to minimize conflicts this should be reviewed and merged only *after* PR 117 has been merged. there will be 20 files changed after that. other than the deprecation code itself, all the changes are to replace `last = Event.last` and `{last} = Event` with `last = ao.lastEvent`, and analogous changes for `Span.last`

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -49,7 +49,7 @@ module.exports = function (appoptics) {
     readyToSample,
     getTraceSettings,
     sampling,
-    stringToMetadata,
+    xtraceToEvent,
 
     // emitter (http) instrumentation
     patchResponse,
@@ -425,7 +425,7 @@ function readyToSample (ms, obj) {
  * @typedef {object} TraceSettings
  * @property {boolean} doSample - the sample decision
  * @property {boolean} doMetrics - the metrics decision
- * @property {Event} metadata - the parent event to get task id from
+ * @property {Event} traceTaskId - the parent event to get task id from
  * @property {boolean} edge - true to edge back to parent op id
  * @property {number} source - the sample decision source
  * @property {number} rate - the sample rate used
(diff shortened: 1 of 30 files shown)


lib/api.js
```diff
@@ -480,19 +476,23 @@ function getTraceSettings (xtrace, options = {}) {
       source: 5,
       rate: 0,
       edge: false,
-      metadata: aob.Event.makeRandom(0)
+      metadata: aob.Event.makeRandom(0),
```
prettycoder, line 479: It looks like `metadata` is still in here to serve a default further down, L484, but i don't see why it needs to be added to osettings.
My proposal would be to to remove metadata here and rewrite line 485:
`osettings.traceTaskId = aob.Event.makeRandom(0)`

lib/probes/http.js
```diff
@@ -6,8 +6,7 @@ const os = require('os')
 const semver = require('semver')
 
 const ao = require('..')
-const Span = ao.Span
-const Event = ao.Event
+const Span = ao.Span;
```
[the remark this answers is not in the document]
bmacnaughton: unless you have a different version it's on line 433. sometimes i miss this kind of thing but not too often; i run vscode with eslint running - it flags an error if there are variables declared but never used (or declared with `let` but never modified.

lib/api.js
bmacnaughton: it needs to be added to osettings so that i don't have to change appoptics-bindings-node in lock step. it returns metadata (and not traceTaskId). the code you're flagging is only executed if there was an error getting settings from oboe. once that oboe has been released i can switch over. but in the meantime i want the agent to operate with or without a compatible version of bindings.

bmacnaughton:
good catch on the api - it generate it automatically from the sources but it does nedd to be updated. i'll add that to a subsequent PR.

lib/probes/http.js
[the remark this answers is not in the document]
prettycoder: right, sorry!

lib/api.js
prettycoder: ok

PR #121, opened 2020-06-26
v8 final changes for rc

bmacnaughton:
- comment out notifications requires/enabling
- pass aob as argument to constructor in notification - keeps bindings loader in one place
- remove span.parent (was only assigned and never read)
- don't send metrics unless span.topSpan
- migrating guide reflects deleted Event.last and Span.last
- notifications tests skip when notifications are not enabled


lib/get-unified-config.js
```diff
@@ -65,7 +65,11 @@ function getUnifiedConfig () {
     reporter: {location: 'b', type: ['ssl', 'udp', 'file']},
     endpoint: {location: 'b', type: 's', name: 'COLLECTOR'},
     trustedPath: {location: 'b', type: 's', name: 'TRUSTEDPATH'},
-
+    unifiedLogging: {
+      location: 'b',
+      type: ['always', 'never', 'preferred'],
+      default: 'preferred'
```
prettycoder, line 71: just curious ... what does 'preferred' mean?

test/probes/http-common.js
```diff
@@ -917,10 +915,13 @@ describe(`probes.${p}`, function () {
           emitter,
           function (done) {
             const req = driver.get(url, function (res) {
+              // when we get data abort the request
```
prettycoder, line 918: I think for future readers of the code it would interesting to know **why** the request is aborted, resp. why there is a delay needed on line 897.
The comments explain what is being done--which can be read from the code--but the why is less obvious.

lib/index.js
```diff
@@ -441,12 +427,34 @@ if (config.traceMode === 0) {
...
+  //
+  if (options.unifiedLogging === 'preferred' && !badVersion || options.unifiedLogging === 'always') {
+    const Notifications = require('./notifications');
+    const notificationListener = require('./notifications-listener');
+    ao.notifications = new Notifications(notificationListener, bindings);
+    ao.notifications.startNotifier();
+  } else {
+    let msg = 'unified logging disabled by configuration';
+    if (options.unifiedLogging === 'preferred') {
+      msg += ` for node ${process.version}`;
```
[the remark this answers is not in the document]
bmacnaughton: maybe i'm missing what you're saying but if `options.unifiedLogging` was `never` it is disabled by config. if it was preferred it will only be disabled when the node version has the RSS problem. that's what i'm trying to do anyway.

lib/get-unified-config.js
bmacnaughton: preferred means to try to use it (prefer to use it). it will be disabled if not compatible with the node version being executed.

test/probes/http-common.js
bmacnaughton: fair comment. this is what i'm working on now and will expand the comment.

bmacnaughton: but i think the why is stated in the test - make sure that an error is not reported when the client aborts a connection. make sense?

PR #122, opened 2020-07-03
Cassandra fixes

bmacnaughton:
- add support for cassandra-driver >= v4.4.0
- document proxy in configuration guide.

### DIFF
--- a/lib/probes/cassandra-driver.js
+++ b/lib/probes/cassandra-driver.js
@@ -174,7 +174,6 @@ function patchClientStream (client, consistencies) {
 }
 
 function patchClientExecute (client, consistencies) {
-  // 4.4.0+ _execute: async ƒ (query, params, execOptions)
   if (typeof client._execute !== 'function' || client._execute.constructor.name !== 'AsyncFunction') {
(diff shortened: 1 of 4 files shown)


lib/probes/cassandra-driver.js
```diff
@@ -155,21 +154,87 @@ function serializeList (list) {
...
+      },
+      () => execute.apply(this, args),
+      conf
+    )
+  })
+}
+
+
 function patchClientInnerExecute (client, consistencies) {
+  // 4.4.0+ _execute: async ƒ (query, params, execOptions)
```
prettycoder, line 237: not sure what this comment helps here, should it be `_innerExecute`

bmacnaughton: nice catch. this was a working note documenting the new signature that i originally made in `patchClientInnerExecute()`. that was copied when i cut/pasted/modified `patchClientInnerExecute()` to create `patchClientExecute()` so it then appeared in both places. i've removed both comments and added a more meaningful comment where the decision on which function to patch is made.

PR #123, opened 2020-07-19
lib/probes: use info.version

bmacnaughton:
require-patch fetches the version; remove version fetches from probes.
- bunyan
- generic-pool
- koa-route
- koa-router
- mongoose
- mysql
- cassandra-driver
- pg
- pino
- raw-body
- winston

### DIFF
--- a/lib/probes/winston.js
+++ b/lib/probes/winston.js
@@ -49,19 +49,16 @@ function patchLog (log) {
 // winston's logger module. but patching logger is triggered by winston being
 // required.
 //
-module.exports = function (winston) {
+module.exports = function (winston, info) {
   if (!ao.probes.winston.enabled) {
     return winston;
   }
 
   requirePatch.disable();
-  let pkg;
   let target;
   try {
-    pkg = requirePatch.relativeRequire('winston/package.json');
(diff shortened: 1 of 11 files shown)


lib/probes/winston.js
```diff
@@ -70,13 +67,13 @@ module.exports = function (winston) {
...
 
-  if (semver.gte(pkg.version, '3.0.0')) {
+  if (semver.gte(info.version, '3.0.0')) {
     if (typeof target.prototype.write === 'function') {
       shimmer.wrap(target.prototype, 'write', patchWrite);
     } else {
       logMissing('Logger.prototype.write()');
     }
-  } else if (semver.gte(pkg.version, '1.0.0') && semver.lt(pkg.version, '3.0.0')) {
+  } else if (semver.gte(info.version, '1.0.0') && semver.lt(info.version, '3.0.0')) {
```
prettycoder, line 76: `&& semver.lt(info.version, '3.0.0')` seems redundant at this point as it won't get here if it is >= 3.0.0

bmacnaughton: good catch.

PR #125, opened 2020-08-19
8.1.0 ClientIP fix

bmacnaughton:
- http server behind nginx ClientIP is missing/not useful
- add configuration option to specify remote address header
- set ClientIP to '' when ClientIP is undefined so Event.set() doesn't log error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "appoptics-apm",
-  "version": "8.0.1",
+  "version": "8.1.0",
   "description": "Agent for AppOptics APM",
   "main": "lib/index.js",
   "bin": {
--- a/test/probes/http-common.js
+++ b/test/probes/http-common.js
@@ -509,7 +509,32 @@ describe(`probes.${p}`, function () {
         }
         request(options);
       });
-    })
+    });
+
+    it('should set ClientIP when req[\'client-ip-header\'] is undefined', function (done) {
+      const server = createServer(options, function (req, res) {
+        res.end('done');
+      });
+
+      helper.doChecks(emitter, [
+        function (msg) {
+          check.server.entry(msg);
+        },
+        function (msg) {
+          check.server.exit(msg);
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "appoptics-apm",
-  "version": "8.0.1",
+  "version": "8.1.0",
```
prettycoder, line 3: why not 8.0.2? 
I seems like a fix to me

test/probes/http-common.js
```diff
@@ -481,6 +481,36 @@ describe(`probes.${p}`, function () {
...
+      server.listen(function () {
+        const port = server.address().port;
+        const options = {
+          url: `${p}://localhost:${port}`,
+          headers: {'x-real-ip': ClientIPExpected}
+        }
+        request(options);
+      });
+    })
+
```
prettycoder, line 513: can you test that it works when the header is undefined?
I think that is the test case we want to make sure doesn't break with future code changes.

package.json
bmacnaughton: i bumped the minor version because i didn't just fix the bug but also added the feature to allow mapping a different header to the ClientIP KV. that's likely to be useful in nginx or other reverse-proxy environments.

test/probes/http-common.js
bmacnaughton: i can't test what i really want to (how does the client handle not having a `socket.remoteAddress` property without spending a lot of time on the test. i'd have to patch the cached http/https modules to get control before the probe sees the `req` object or i'd have to figure out what's going on with nginx/nodejs that causes `remoteAddress` to be undefined - i'm not sure how the TCP protocol can fail to include a source IP address.

i added a specific test as a placeholder for a more sophisticated test in the future. currently all tests verify that ClientIP is present in the KV pairs when 'client-ip-header' is falsey.

PR #127, opened 2020-11-05
additional lambda changes

bmacnaughton:
the bulk of the non-test changes are to use a native promise interface to execute lambda functions. previously `pStartOrContinueTrace` was a simple wrapper that wrapped the user's promise-returning function and then called `startOrContinueTrace` which used the existing callback mechanism to signal completion. that arrangement didn't allow the result of the promise to be evaluated before exiting the span, and that is required in the lambda environment in order to detect non-error returns that the agent should treat as errors. `pStartOrContinueTrace` now parallels `startOrContinueTrace` with a few optimizations. a related change is a new `Span.runPromise()` instance function.

the agent now captures whether an inbound-x-trace was received; lambda functions do not attempt to return an x-trace unless one was received. a fair amount of code was commented out that would insert a response x-trace header even when the user did not insert their own headers. now, if there is not a headers object in the response, an x-trace header will not be inserted. previously the agent would modify the response to insert a header object if necessary.

update to test.sh to allow skipping groups, "only" groups, and add color output.

about half the lines of the changes are to testing.
- axios updates making tests complete on promises, not callbacks
- a standalone test that doesn't use mocha as i was tracking down "why is context hanging around?"
- custom test had dramatic changes (promise changes, log message changes, should => expect)
- helper.js got a few more options for debugging testing.
- create module to compose the lambda version from another module's package.json.
- changes to lambda local and remote tests for changes in x-trace header insertion and error handling

@cheempz - including you as you might want to check out `guides/lambda-configuration.md`

### DIFF
--- a/guides/api.md
+++ b/guides/api.md
@@ -31,7 +31,6 @@
 
 * [ao](#ao)
     * [.logLevel](#ao.logLevel)
-    * [.serviceKey](#ao.serviceKey)
     * [.loggers](#ao.loggers)
     * [.traceMode](#ao.traceMode)
     * [.tracing](#ao.tracing)
@@ -56,6 +55,7 @@
     * [.sendMetrics(metrics, [gopts])](#ao.sendMetrics) ⇒ [<code>SendMetricsReturn</code>](#SendMetricsReturn)
     * [.getFormattedTraceId()](#ao.getFormattedTraceId) ⇒ <code>string</code>
     * [.insertLogObject([object])](#ao.insertLogObject) ⇒ <code>object</code>
+    * [.wrapLambdaHandler([handler])](#ao.wrapLambdaHandler) ⇒ <code>function</code>
 
 <a name="ao.logLevel"></a>
 
@@ -75,16 +75,6 @@ ao.logLevel = 'warn,error'
 ```js
 var settings = ao.logLevel
 ```
-<a name="ao.serviceKey"></a>
-
-### ao.serviceKey
-**Kind**: static property of [<code>ao</code>](#ao)  
-**Properties**
-
-| Type | Description |
-| --- | --- |
-| <code>string</code> | the service key |
-
 <a name="ao.loggers"></a>
 
 ### ao.loggers
@@ -674,6 +664,32 @@ logger.info(ao.insertLogObject({
     more: 'there will be an added ao property'
 }))
 ```
+<a name="ao.wrapLambdaHandler"></a>
+
+### ao.wrapLambdaHandler([handler]) ⇒ <code>function</code>
+Wrap the lambda handler async function so it can be traced by AppOptics APM.
+
+**Kind**: static method of [<code>ao</code>](#ao)  
+**Returns**: <code>function</code> - - an async function, wrapping handler, to be used
+                      instead of handler.  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| [handler] | <code>function</code> | wraps your lambda handler function so it                               is instrumented. your handler must return                               a promise if it is not an async function. |
+
+**Example**  
+```js
+const ao = require('appoptics-apm');
+
+const wrappedHandler = ao.wrapLambdaHandler(myHandler);
+
+async function myHandler(event, context) {
+  // implementation
+}
+
+// set the handler to the wrapped function.
+exports.handler = wrappedHandler;
+```
 <a name="Span"></a>
 
 ## Span
@@ -684,6 +700,7 @@ logger.info(ao.insertLogObject({
     * _instance_
         * [.descend(name, data)](#Span+descend) ⇒ [<code>Span</code>](#Span)
         * [.run(fn)](#Span+run) ⇒
+        * [.runPromise(fn, [opts])](#Span+runPromise) ⇒
         * [.runAsync(fn)](#Span+runAsync) ⇒
         * [.runSync(fn)](#Span+runSync) ⇒
         * [.enter(data)](#Span+enter)
@@ -758,6 +775,26 @@ span.run(function () {
 span.run(function (wrap) {
   asyncCallToTrace(wrap(callback))
 })
+```
+<a name="Span+runPromise"></a>
+
+### span.runPromise(fn, [opts]) ⇒
+Run an promise-returning function within the context of this span.
+
+**Kind**: instance method of [<code>Span</code>](#Span)  
+**Returns**: the value returned by fn()  
+
+| Param | Type | Default | Description |
+| --- | --- | --- | --- |
+| fn | <code>function</code> |  | the promise-returning function to run within the span context |
+| [opts] | <code>object</code> |  | an options object |
+| [opts.collectBacktraces] | <code>boolean</code> | <code>false</code> | add a backtrace KV to the span |
+
+**Example**  
+```js
+span.runPromise(async function () {
+  return axios.get('https://google.com');
+});
 ```
 <a name="Span+runAsync"></a>
 
--- a/guides/api.md
+++ b/guides/api.md
@@ -667,7 +667,7 @@ logger.info(ao.insertLogObject({
 <a name="ao.wrapLambdaHandler"></a>
 
 ### ao.wrapLambdaHandler([handler]) ⇒ <code>function</code>
-Wrap the lambda handler async function so it can be traced by AppOptics APM.
+Wrap the lambda handler function so it can be traced by AppOptics APM.
 
 **Kind**: static method of [<code>ao</code>](#ao)  
 **Returns**: <code>function</code> - - an async function, wrapping handler, to be used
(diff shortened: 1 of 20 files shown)


lib/api.js
```diff
@@ -1090,6 +1097,7 @@ function startOrContinueTrace (xtrace, build, run, opts, cb) {
  * @param {boolean} [opts.collectBacktraces=false] - collect backtraces
  * @param {boolean} [opts.forceNewTrace=false] - ignore any existing context and force a new trace
  * @param {string|function} [opts.customTxName] - name or function
```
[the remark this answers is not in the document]
bmacnaughton: it's more of a jsdoc syntax question. the last params is an optional options object. if present the object can contain properties, each of which is documented as `opts.propertyName`. so they aren't really separate params but are the last params expanded.

lib/span.js
```diff
@@ -223,6 +226,132 @@ Span.prototype.run = function (fn) {
   return fn.length === 1 ? this.runAsync(fn) : this.runSync(fn)
 }
 
+Span.prototype.isLambdaTopSpan = function () {
+  return typeof this.topSpan === 'string';
```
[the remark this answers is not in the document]
bmacnaughton: it's explained in the place where it is set. you asked about this the last time but now that i decided to start isolating that interpretation to a single place i should probably explain it there. good point.

lib/api.js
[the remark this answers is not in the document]
prettycoder: I could have figured that out, if I had paid better attention :(

lib/span.js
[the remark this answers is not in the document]
prettycoder: ooops, sorry my memory isn't that great

lib/span.js
```diff
@@ -262,25 +391,25 @@ Span.prototype.runAsync = function (fn) {
...
-  const ret = fn.call(span, (cb, handler) => ao.bind(function (err) {
+  const ret = fn.call(span, (cb, handler) => ao.bind(function (error) {
     if (handler) {
       // handler is present only for some memcached functions.
       // TODO BAM how to handle this with customTxName...
       handler.apply(this, arguments)
     } else {
       // this is the "normal", i.e., non-memcached path.
-
+      const et = (Date.now() - startTime) * 1000;
```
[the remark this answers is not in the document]
bmacnaughton: funny - it moved because an intermediate approach to solving the span exit required it. i just left it there but probably shouldn't because it creates a bit of extra work for regular spans.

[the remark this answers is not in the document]
bmacnaughton: addressed

lib/span.js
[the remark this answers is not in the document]
bmacnaughton: addressed

prettycoder (approved):
looks good to me!

guides/api.md
```diff
@@ -31,7 +31,6 @@
 
 * [ao](#ao)
     * [.logLevel](#ao.logLevel)
-    * [.serviceKey](#ao.serviceKey)
```
cheempz, line 34: Is `serviceKey` no longer supported?

guides/api.md
```diff
@@ -674,6 +664,32 @@ logger.info(ao.insertLogObject({
     more: 'there will be an added ao property'
 }))
 ```
+<a name="ao.wrapLambdaHandler"></a>
+
+### ao.wrapLambdaHandler([handler]) ⇒ <code>function</code>
+Wrap the lambda handler async function so it can be traced by AppOptics APM.
```
cheempz, line 670: Hmm I had the impression that you got the wrapper to work with both the "async function"-based AWS lambda handler, and the "callback"-based handler... but now reading above and the KB draft, is that true or does it only support the async-function-based handler?

guides/api.md
```diff
@@ -674,6 +664,32 @@ logger.info(ao.insertLogObject({
...
+
+**Kind**: static method of [<code>ao</code>](#ao)  
+**Returns**: <code>function</code> - - an async function, wrapping handler, to be used
+                      instead of handler.  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| [handler] | <code>function</code> | wraps your lambda handler function so it                               is instrumented. your handler must return                               a promise if it is not an async function. |
```
cheempz, line 678: Maybe related to above comment, not sure if:
> your handler must return a promise if it is not an async function.
the "not async function" means the same thing as the "non-async handler" described in https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html... in which case do we only wrap if the customer's "non-async handler" returns a promise?

guides/api.md
bmacnaughton: no, it's not. it's not relevant in a lambda environment and it's not at all clear why a program would ever need it.

guides/api.md
bmacnaughton: it works with both.

guides/api.md
bmacnaughton: that's actually impossible to know - what will the user's handler return?

we wrap the handler and allow it to complete via a promise or a callback. see `promiseOnlyHandler()`

guides/lambda-configuration.md
```diff
@@ -3,15 +3,20 @@
...
+
+If you haven't already, set up the forwarder. This only needs to be done once per region
+for the account.
 
 Add the `appoptics-node` layer to your function
-- screen shots?
-- select `arn:aws:lambda:${your-function's-region}:858939916050:layer:appoptics-node`:41
-- forwarder setup already done (where to document this - each agent?)
+- find the current version for your region(s) at https://files.appoptics.com/lambda-layers/node/
+- copy the arn (e.g., `arn:aws:lambda:us-east-1:085151004374:layer:appoptics-node:7`)
```
cheempz, line 13: In the KB docs draft for the [Node layer](https://docs.google.com/document/d/1s63i3_NZn-H6r7Wyb4ObercjWxL3292dbkCeWJO3dR4/edit#heading=h.lddh8ahnezo2) we're calling it `appoptics-node`, with the index file located under https://files.appoptics.com/lambda-layers/appoptics-node 

Also clicking the file for us-east-1 (https://files.appoptics.com/lambda-layers/node/us-east-1) downloaded rather than displayed it, is there some way to get it to display like https://files.appoptics.com/lambda-layers/appoptics-apm-layer instead?

And noticed that the downloaded file for us-east-1 has the layer name as `appoptics-apm-layer` instead of `appoptics-node`
```arn:aws:lambda:us-east-1:085151004374:layer:appoptics-apm-layer:5 - apm v9.0.0-lambda-27, bindings v10.0.0-lambda-8, auto v1.0.1-lambda-3```

File: guides/lambda-configuration.md [hunk 8]
```diff
@@ -3,15 +3,20 @@
 This covers how to configure and enable AppOptics APM for nodejs lambda functions. For
 general configuration options see `configuration.md` in the same directory.
 
-## Prerequisite
+## Prerequisites
+
+If you haven't already, set up the forwarder. This only needs to be done once per region
+for the account.
```
cheempz, line 9: Going back to the question you raised at standup -- there will be an "overview" section in the KB docs about our lambda support and a section on the [forwarder](https://docs.google.com/document/d/1s63i3_NZn-H6r7Wyb4ObercjWxL3292dbkCeWJO3dR4/edit#heading=h.34fnvwia9xym) that you can link to from here... as soon as the KB draft is put in staging ;)

guides/api.md
cheempz: OK so why the `your handler must return a promise`?

guides/api.md
cheempz: Ah good -- I find the "lambda handler async function" confusing then, how about just saying "lambda handler function" in line 670 then.

guides/lambda-configuration.md
bmacnaughton: Got that. The rest of the doc reads OK to you?

guides/lambda-configuration.md
bmacnaughton: > with the index file located under https://files.appoptics.com/lambda-layers/node. 

this is just a test. the name can be anything desired. nice catch in the docs.

> is there some way to get it to display like https://files.appoptics.com/lambda-layers/appoptics-apm-layer instead?

i was originally going to ask you about this. it turns out that making files appear requires running the `update-prod-index` jenkins job, making sure they have correct permissions is another step (need to add ACL public-read), and whether the file is downloadable or not is unknown to me at this time; i don't have console access to the production s3 bucket so it is a black box. the reason that the files were published under node is that i didn't know whether we had permissions to delete keys or not and i didn't want to create any that couldn't be deleted; it turns out i can delete keys so it wasn't an issue.

so hopefully you know why it wants to download the file. looking at each using `aws s3api` they are both text/plain at this time and yet us-west-2 displays while us-east-1 offers a download.

> And noticed that the downloaded file for us-east-1 has the layer name as appoptics-apm-layer

this is because we don't want to publish any `appoptics-node` layers in production until i publish a real one. publishing `appoptics-node` to production will use up version numbers in the production layer, so these files were run against a different layer name to test. i can publish appoptics-node layers in the apm-instrumentation account and write those to the production files area if you'd like to see that.

guides/api.md
bmacnaughton: mostly because i wasn't focused on documenting what seems like a legacy, though still supported, interface. am changing to make it more clear.

guides/lambda-configuration.md
cheempz: Yep looks good!

guides/lambda-configuration.md
cheempz: Looks like the display is working now 👍

guides/api.md
bmacnaughton: discussed in slack

cheempz (approved):
LGTM, thanks @bmacnaughton !

PR #131, opened 2020-11-25
Lambda fixes

bmacnaughton:
- flag non-api-gateway/rest lambda invocations as lambda
- report token bucket KVs as strings
- fix unrelated-to-lambda transaction name bug

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -1701,9 +1701,9 @@ function getTraceContext (headers) {
   return undefined;
 }
 
-function makeLambdaKVPairs (spanType, event, context) {
+function makeLambdaKVPairs (topSpanType, event, context) {
   const kvpairs = {
-    Spec: typeof spanType === 'string' ? 'aws-lambda:ws' : 'aws-lambda',
+    Spec: topSpanType !== 'lambda-generic' ? 'aws-lambda:ws' : 'aws-lambda',
     InvocationCount: ao.lambda.invocations,
   };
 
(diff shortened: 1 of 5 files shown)


lib/api.js
```diff
@@ -1614,9 +1616,6 @@ function wrapLambdaHandler (userHandler) {
       finalize = (span) => {
         span.defaultTxName = process.env.APPOPTICS_TRANSACTION_NAME || `${method.toUpperCase()}.${userHandlerName}`;
```
cheempz, line 1617: Would we end up with a transaction name like `.handlerName` if there is no method?

lib/span.js
```diff
@@ -125,7 +125,7 @@ Span.makeEntrySpan = function makeEntrySpan (name, settings, kvpairs) {
   // supply a default in case the user didn't provide a txname string or
   // function that returns a string. if the span is unnamed then let oboe
   // provide "unknown". there is no customTxName function by default.
-  span.defaultTxName = span.name ? 'custom-' + span.name : '';
+  span.defaultTxName = span.name ? 'default-' + span.name : '';
```
cheempz, line 128: Minor: we actually sort of agreed in https://swicloud.atlassian.net/browse/AO-8450 that for SDK started traces we'll just default to something like `custom-<top span name>`... does the above line impact that?

lib/probes/http.js
```diff
@@ -566,7 +566,7 @@ function patchServer (module, options, protocol) {
       if (span.doMetrics) {
 
         // set this value only if doing metrics.
-        exitKeyValuePairs.TransactionName = res._ao_metrics.txName
+        exitKeyValuePairs.TransactionName = res._ao_metrics.txname;
```
cheempz, line 569: Do we need to make a regular (non-lambda) agent release to get this fix in?

cheempz (approved):
LGTM, thanks @bmacnaughton!  Left some comments but no showstoppers 👍

lib/api.js
bmacnaughton: it will end up as `undefined.handlerName`. thoughts?

lib/span.js
bmacnaughton: yes, let me change it back. trying to clean up something that shouldn't have been.

lib/probes/http.js
bmacnaughton: yes. that release will be based on this lambda code base.

lib/api.js
bmacnaughton: fixed this and now uses lambda-function name, not handler function name.

PR #133, opened 2020-12-07
koa fix and patch optimization

bmacnaughton:
- pass constructor options for koa
- minimize builtin lookups when patching

### DIFF
--- a/lib/probes/koa.js
+++ b/lib/probes/koa.js
@@ -1,45 +1,44 @@
 'use strict'
 
-const shimmer = require('ximmer')
-const Span = require('../span')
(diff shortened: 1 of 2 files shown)


bmacnaughton:
- the koa change is really two lines - pass the option argument that may be supplied when invoking the constructor. all the other changes are just adding semicolons to the end of lines.
- the require-patch change avoids doing a builtin lookup and an attempt to read package.json if not builtin when the module 1) is not instrumented (doesn't have a probe) or 2) has already been patched.

prettycoder (approved):
lgtm

Issue #65, opened 2019-04-30
Ignore certain error types 


martin-rog001:
Further to my conversation with AppOptics support staff, we were discussing potential options to disable certain errors from being recorded by the agent.

I'm currently having an issue where my exception logs are full of thousands of these exceptions:
```
ENOENT: no such file or directory, stat 
```

They aren't of any use to us, so it would be great to have an option to filter these out of the dashboard (https://my.appoptics.com/apm/ - > Exceptions). Or more effectively, tell the agent to ignore these before pushing up to AppOptiocs.

Many thanks

Martin

bmacnaughton:
Martin, thanks for this request. It's implemented in v6.9.0. Please let me know how it works for you.

Take a look at the [guide](https://github.com/appoptics/appoptics-apm-node/blob/master/guides/configuration.md) for some more information but the general idea is:

```
{
  "enabled": true,
  "serviceKey": "...",
  "probes": {
    "fs": {
      "enabled": false,
      "ignoreErrors": {
        "open": {
          "ENOENT": true
        }
      }
    }
  }
}
```

bmacnaughton:
Closing as the feature has been added.

immohitrathi:
I want to ignore below such errors in appoptics which are showing as exceptions for each request, this is non breaking and page works just fine. 

ENOENT: no such file or directory, stat '/opt/node_app/app/public/healthcheck'
ENOENT: no such file or directory, stat '/opt/node_app/app/public/akansha-deep-heights-kunhari-kota-npxid-r353279'
ENOENT: no such file or directory, stat '/opt/node_app/app/public/reviews/project-and-locality-reviews'

As per above comment, should I keep this as it is or I need to pass `enabled` as `true` for `fs` probe in below json config??  I don't want to ignore other `fs` exceptions.

{
  "enabled": true,
  "serviceKey": "...",
  "probes": {
    "fs": {
      "enabled": false,
      "ignoreErrors": {
        "open": {
          "ENOENT": true
        }
      }
    }
  }
}

bmacnaughton:
if `fs` is set `"enabled": false` then no `fs` events will be reported. that will suppress errors but it will also suppress all `fs` events.

i think you want `"enabled": true` if i understand correctly.

immohitrathi:
@bmacnaughton  
I would like to thank you for being so helpful with issues being faced. 

[...]

PR #136, opened 2021-01-23
extend ignore errors to all fs path-based operations

bmacnaughton:
Of the synchronous path-based fs operations, only `openSync` checked for ignored operations.

My thinking was to not introduce even slight additional overhead for the synchronous functions unless it addressed the specific customer problem. while the asynchronous functions needed the finalize function in order to capture the created span, the synchronous functions didn't, so avoiding the finalize function slightly reduced the overhead. in retrospect the slight reduction was not worth 1) different behavior for async and sync operations or 2) avoiding the slight increase in overhead for synchronous operations.

this pr makes path-based operations work the same for async and sync versions - errors from any operation can be ignored. this addresses issue #65 (reopened) and [AO-18483].

[AO-18483]: https://swicloud.atlassian.net/browse/AO-18483

### DIFF
--- a/lib/probes/fs.js
+++ b/lib/probes/fs.js
@@ -176,16 +176,14 @@ function patchPathMethod (fs, method) {
             FilePath: args[0]
           },
         }
-        if (method === 'open') {
-          spanInfo.finalize = function (createdSpan) {
-            span = createdSpan;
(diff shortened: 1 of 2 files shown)


prettycoder (approved):
lgtm :)

PR #140, opened 2021-02-09
Docs updates

cheempz:
* mention suppressing agent messages via APPOPTICS_LOG_SETTINGS from v10.0.0 release
* mention custom metrics are aggregated for AO-18250

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -1335,7 +1335,7 @@ function sendMetric (name, options) {
  * 2) value-based - a specific value (or sum of values if count > 1).
  *
  * The metrics submitted are aggregated by metric name and tag(s), then
- * sent on interval every 60 seconds.
+ * sent every 60 seconds.
  *
  * @method ao.sendMetrics
  * @param {(metric|metric[])} metrics - a metric or an array of metrics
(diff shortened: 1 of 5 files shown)


prettycoder (approved):
lgtm :)

cheempz:
LOL, I realized I was happily running `npm docs` per the contribution notes but it wasn't actually generating the documentation that should have been updated based on the code comment changes!

Looked into this a bit and we should be using `npm run "script action"` so updated the instructions and renamed the script action to be a bit more clear.

Please re-review @prettycoder :)

lib/api.js
```diff
@@ -1334,6 +1334,8 @@ function sendMetric (name, options) {
  *                  value is associated with this type)
  * 2) value-based - a specific value (or sum of values if count > 1).
  *
+ * The metrics submitted are aggregated by metric name and tag(s), then
+ * sent on interval every 60 seconds.
```
prettycoder, line 1338: `on interval` needed? I think the sentence would be fine just saying "The metrics submitted are aggregated by metric name and tag(s), then sent every 60 seconds."

prettycoder (approved):
lgtm :)
minor suggestion

lib/api.js
cheempz: That's true!  Updated.

PR #143, opened 2021-03-30
AO-19439 fix AWS Lambda trace data truncation

cheempz:
This PR updates the agent, when running in AWS Lambda, to clear the nonblocking flag on stdout via a new reporter initialization setting in the bindings package.  Doing so prevents data written to stdout by the agent (and the lambda function code itself) from being truncated/interleaved in the resulting CloudWatch logs. 

The corresponding bindings changes are in https://github.com/appoptics/appoptics-bindings-node/pull/40.


lib/get-unified-config.js
```diff
@@ -260,6 +261,11 @@ function getUnifiedConfig () {
     delete results.global.samplePercent;
   }
 
+  // remove this setting unless we're in the lambda environment
```
bmacnaughton, line 264: rather than do this you probably want to set `eeid: 'lambda'` in the definition. that will cause it to omit the setting if the execution environment is not lambda (as the other lambda-only settings).

cheempz: That's what I tried first in 9be515468d9889c53114656bf76d7cd3676b9602, unfortunately if a setting has the `default` property set, it will stick around.

bmacnaughton: invert the setting? make it non-blocking?

bmacnaughton: obviously your call, but it's a bug and i thought i'd offer a workaround. eeid should not set a default.

prettycoder (approved):
Is stdoutClearNonblocking used anywhere beyond testing?

cheempz:
> Is stdoutClearNonblocking used anywhere beyond testing?

It's used by the bindings, which passes it in during reporter initialization -- it will be set true (1) only if running in the lambda environment.

PR #144, opened 2021-04-02
handle defaults in one place, correctly

bmacnaughton:
set defaults in one place, at the start.

add tests, including a test key which can be replaced with a real one when ready. the test key must have eeid 'lambda' and a default value.

### DIFF
--- a/lib/get-unified-config.js
+++ b/lib/get-unified-config.js
@@ -67,6 +67,7 @@ function getUnifiedConfig () {
     tokenBucketCapacity: {location: 'e', type: 'i', eeid: 'lambda'},
     tokenBucketRate: {location: 'e', type: 'i', eeid: 'lambda'},
     transactionName: {location: 'e', type: 's', eeid: 'lambda'},
+    testKeyWithDefault: {location: 'b', type: 'i', eeid: 'lambda', default: 42},
     // end-user but ignore as it is only used when configuring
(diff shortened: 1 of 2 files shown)


lib/get-unified-config.js
```diff
@@ -67,6 +67,7 @@ function getUnifiedConfig () {
     tokenBucketCapacity: {location: 'e', type: 'i', eeid: 'lambda'},
     tokenBucketRate: {location: 'e', type: 'i', eeid: 'lambda'},
     transactionName: {location: 'e', type: 's', eeid: 'lambda'},
+    testKeyWithDefault: {location: 'b', type: 'i', eeid: 'lambda', default: 42},
```
cheempz, line 70: Should this be part of test setup, instead of an actually defined setting?

bmacnaughton: ideally it would be part of a test setup, but that would require exposing the configuration object as opposed to keeping it isolated to one module. my thought was that you'd replace this config item with `stdoutConfigNonBlocking` and change the test so it used that. i viewed `testKeyWithDefault` as a placeholder.

if that doesn't make sense to you i can expose the config itself and make those changes.

cheempz: I see, yes that works.

cheempz:
Thanks @bmacnaughton!

PR #145, opened 2021-07-28
First take on a dev setup

ronilan:
This pull request creates a Mac dev environment.

1. Modifies `.gitignore` (a6f3fe2).
2. Adds Mocha as local dev dependency (af24dfd).
3. Everything related to dev is in dev directory (ef8d259)
  - Dev container has build tools, docker, docker compose and multiple versions of node installed as well as specific Oracle and PostgreSQL related installs.
 - The dev container is setup with multiple connected services using docker compose.
 - There is simple way to get local "one off" containers.
 - There is an easy way to work with "formalized" dev reop.

4. Four npm commands launch four shell scripts (e53e65d):
a. `npm run dev` - starts and ubuntu based dev environment defined in a docker compose. Code is mounted. This is where work should be done.
b. `npm run dev:oneoff` - starts a "one off" dev environment. code is copied. This is for testing specific node/os things.
c. `npm run dev:repo:reset` - runs a set of git operations that "resets" the dev repo to match the main repo.
d. `npm run dev:reinstall` - removed node_modules and runs npm install. To be used after switching node versions.

The setup is similar to the one used in upstream repo https://github.com/appoptics/appoptics-bindings-node/pull/63


.gitignore
```diff
@@ -1,5 +1,32 @@
...
-node_modules
+# anything, anywhere starting with __ or ~~
+__*
+~~*
+
+# Local development
+*.env
+*.dev
+
+# Apple
```
prettycoder, line 9: I recently discovered global gitignore https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer
imho it is a cleaner way to exclude machine specific files

dev/Dockerfile
```diff
@@ -0,0 +1,60 @@
+FROM ubuntu:20.04
+SHELL ["/bin/bash", "-c"]
+
+# tools
+RUN su - \
```
prettycoder, line 5: I never need su or sudo on an ubuntu image, why do you need it?

dev/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+                g++ \
+                python \
+                make \
+                postgresql-server-dev-12 \
+                zip \
+                unzip
+
+# install docker (inception)
+RUN curl -fsSL https://get.docker.com -o get-docker.sh && \
+    sudo sh get-docker.sh
```
prettycoder, line 21: sudo needed?

dev/repo/reset.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
prettycoder, line 1: interesting ... why is this needed?

package.json
```diff
@@ -14,7 +14,11 @@
...
     "gen-docs": "cat lib/index.js lib/api.js lib/span.js lib/event.js > api-docs-temp.js && jsdoc2md -f api-docs-temp.js > guides/api.md && rm api-docs-temp.js",
     "install-deps": "./deps.sh",
-    "lint": "eslint --quiet lib/*.js lib/**/*.js test/*.js test/**/*.js prepack.js"
+    "lint": "eslint --quiet lib/*.js lib/**/*.js test/*.js test/**/*.js prepack.js",
+    "dev": "dev/start.sh",
+    "dev:oneoff": "dev/oneoff.sh",
+    "dev:repo:reset": "dev/repo/reset.sh",
+    "dev:reinstall": "rm -rf node_modules && npm install --unsafe-perm"
   },
   "author": "Bruce A. MacNaughton <bruce@solarwinds.cloud>",
```
prettycoder, line 23: do you wanna add your name?

prettycoder (approved):
lgtm

just one question about sudo/su in the Dockerfile

package.json
ronilan: A lot of config mods incoming... will add it sometime...

dev/repo/reset.sh
ronilan: See other repo.

dev/Dockerfile
ronilan: Naively taken from [docs](https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script). Most likely not needed.

Thinking things over, since currently, neither Docker nor Docker Compose are needed, and since there are other problems with such inception, the whole thing was removed. 

Re-review will be needed (repo setting nullify reviews on code change...).

.gitignore
ronilan: I'm actually pulling to the other direction (for now), avoiding anything global (e.g. packages). The goal is to get to a point where node development can be done on a factory reset Mac with minimal configuration (git + docker +  ??? ). There is some more WIP here as there are more repos involved (testing).

dev/Dockerfile
ronilan: removed.

prettycoder (approved):
looks good

PR #147, opened 2021-08-03
Dev setup hardening

ronilan:
This pull request adds guard conditional around force push into dev repo.

Note that main repo also has branch protection on master.

### DIFF
--- a/dev/repo/reset.sh
+++ b/dev/repo/reset.sh
@@ -1,11 +1,13 @@
 #!/bin/bash
 
-# script used to clone main repo into dev repo.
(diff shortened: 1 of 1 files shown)


cheempz (approved):
LGTM, thanks @ronilan!

cheempz (changes requested):
Oops, actually took another look and I'm still seeing the force push in an earlier part of the script line 25... that was the part causing my paranoia.

cheempz (approved):
LGTM, thanks @ronilan!

PR #149, opened 2021-08-13
Dev setup Update in Prep for GitHub Actions

ronilan:
This pull request further simplifies the dev setup configuring it to match (as much as possible) what will be used for GitHub Actions

Main changes include:
1. Adding GitHub CLI to the main Dockerfile
2. Removing all service Dockerfiles. All containers use an image pulled from elsewhere eliminating need for build from Docker files.

The setup is also in sync with to the one used in upstream repo https://github.com/appoptics/appoptics-bindings-node/pull/69


prettycoder (approved):
lgtm

PR #150, opened 2021-08-16
AO-20279-Update-NPM-Build-Test-Setup-2

ronilan:
This pull request is the **2nd** in a series of pull requests to update the test and build setup and manage dependencies in package.json.

It includes an additional minimal set of tweaks required to allow upcoming GithHub Actions test and release workflows to work.

First commit (0df5e9c) commented out "flaky" test that fails only when running on GitHub actions runner but passes locally (ubuntu and alpine). A TODO was added.

Second commit (15f03b7) modified the test runner script (`test.sh`) to accept a test group name parameter so that it can run specific subset of tests via command line. The setup is required for GitHub Actions. Bypassing the runner script by calling mocha directly with `npx` currently leads to additional failures.

Third commit (a912a7f) replaces `.npmignore` with files entry in package.json. Generally, explicit include (via files) works better than explicit exclude (`.npmignore`) in the context of npm. It also changed: a) number of files included in package from 68 to 62 removing: 1 accidentally included file and 5 guide files aimed at developers, b) removed (for now) the contributing section of the readme.

Last commit (dac4b15) removes outdated checks from the prepack script.


prepack.js
```diff
@@ -1,34 +1,11 @@
 #!/usr/bin/env node
 'use strict';
 
-// add this check because withFileTypes was added in 10.10.0. for simplicity
-// just require node 12.
-if (process.versions.node.split('.').shift() < 12) {
```
prettycoder, line 6: have we excluded Node 10.9 and earlier somewhere else
the code using withFileTypes is still on line 89, 94 ...

prepack.js
```diff
@@ -1,34 +1,11 @@
...
-}
-
 const fs = require('fs');
-
 const files = fs.readdirSync('.', 'utf8');
 
 let errorCount = 0;
 
-//
-// make sure that there are no ubuntu-* files left around
```
prettycoder, line 19: where did these come from?
I guess it was some docker artifact and you're not creating them anymore

prepack.js
ronilan: The script runs as part of the publish process. It guards from human error, which in this case was publishing from lower version of node (that does not support the `withFileTypes` settings). Moving forward publish will always run on GH Actions, thus error not possible and check not needed.

prepack.js
ronilan: `ubuntu-*` files were (are) the output from running `testeachversion`. We no longer need to test this via script because a) we use an explicit file list and b) we will be running this in GH Actions moving forward.

There may have been also other "generators" of such files but they are no longer present in our setup.

prettycoder (approved):
thanks for the answers
lgtm

PR #164, opened 2021-09-10
AO-20279-Update-NPM-Build-Test-Setup-6

ronilan:
This Pull Request is the **6th** in a series of pull requests to update the build setup and manage dependencies in package.json.

Linting was ran on JavaScript, YAML and JSON files using new implemented configuration in PR #163

This linting pass should **not** be considered a code change. 

From manual inspection looks like auto fixing did only three things: removed semicolons, removed dangling commas and added white-space to objects and comments. Where auto fixing did not change JavaScript code, and an eslint error was emitted using the new configuration code was changed or a comment was made to disable eslint. However all code changes do not impact the library code itself, these will be included in another PR.

All tests pass.

PR #163, opened 2021-09-09
AO-20279-Update-NPM-Build-Test-Setup-5

ronilan:
This Pull Request is the **5th** in a series of pull requests to update the build setup and manage dependencies in package.json. 

It updated eslint setup. 

Now using the @standard style guide and able to lint YAML with https://github.com/aminya/eslint-plugin-yaml and json with https://github.com/azeemba/eslint-plugin-json.

PR #165, opened 2021-09-10
AO-20279-Update-NPM-Build-Test-Setup-7

ronilan:
This Pull Request is the **7th** in a series of pull requests to update the build setup and manage dependencies in package.json.

Core library files were linted and auto-fixed in PR #164 using new implemented configuration in PR #163. This pull request handles the cases where auto fixing did not change JavaScript code, and an eslint error was emitted.

This linting pass should **not** be considered a code change. 

Manual edits included:
1. Changes to conditionals ([no-mixed-operators](https://eslint.org/docs/rules/no-mixed-operators))
2. Removing unneeded constructors ([no-useless-constructor](https://eslint.org/docs/rules/no-useless-constructor))
3. Modifying how existence of object own properties are evaluated ([no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins))

Disabling eslint was added for:
1. Reassigning values in return statement ([no-return-assign](https://eslint.org/docs/rules/no-return-assign))
2. Return in async ([no-unsafe-finally](https://eslint.org/docs/rules/no-unsafe-finally)). (revisit TODO added).

All tests pass.

### DIFF
--- a/lib/metrics.js
+++ b/lib/metrics.js
@@ -24,7 +24,7 @@ class Metrics {
       // don't let metrics keep the process alive
       this.id.unref()
     }
-    return this.state = 'started'
+    return this.state = 'started' // eslint-disable-line no-return-assign
   }
 
   stop () {
@@ -33,7 +33,7 @@ class Metrics {
     }
     aob.metrics.stop()
     clearInterval(this.id)
-    return this.state = 'stopped'
+    return this.state = 'stopped' // eslint-disable-line no-return-assign
   }
 
   getState () {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,7 +16,7 @@ module.exports = {
   rules: {
     'node/handle-callback-err': 'warn',
     'node/no-deprecated-api': 'warn',
-    'camelcase': 'warn',
+    camelcase: 'warn',
     'no-eval': 'warn'
   }
 }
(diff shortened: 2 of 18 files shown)


lib/metrics.js
```diff
@@ -24,7 +24,7 @@ class Metrics {
       // don't let metrics keep the process alive
       this.id.unref()
     }
-    return this.state = 'started'
+    return this.state = 'started' // eslint-disable-line no-return-assign
```
cheempz, line 27: Minor and maybe stylistic: since it's deemed not as readable, can this be turned into two statements, an actual assign followed by a return?

lib/metrics.js
```diff
@@ -33,7 +33,7 @@ class Metrics {
     }
     aob.metrics.stop()
     clearInterval(this.id)
-    return this.state = 'stopped'
+    return this.state = 'stopped' // eslint-disable-line no-return-assign
```
cheempz, line 36: Same question as above

cheempz (approved):
LGTM, thanks @ronilan!  That was a good excuse for me to learn a bit about javascript:)  Good call on revisiting the finally logic in `lib/span.js`, I think it is necessary contortions to deal with AWS Lambda function return values but definitely can use more clarity.

lib/metrics.js
ronilan: Yep. But I'd rather not touch *any* of the reassigns in this pass.

PR #187, opened 2021-11-23
AO-20643-Verify-Workflow-flaky-due-to-AppOptics-not-ready-to-sample

ronilan:
This pull request applies the simplest solution to the "flakiness" problem - allow oboe to try by itself - it increased readyToSample wait five fold to 5000 ms.


prettycoder (approved):
again ;)

PR #189, opened 2021-11-24
AO-16676-Replace-request-with-axios

ronilan:
This pull request replaces `request` with `axios` in tests as `request` is now fully deprecated.

Note that `request` itself is still instrumented and hence there is a `request` test and it is still a "dev dependency".


cheempz (approved):
LGTM, thanks @ronilan

PR #197, opened 2021-12-02
NH-5262-Get-hapi-Components-from-@hapi 

ronilan:
### Overview:
This Pull request removes support from old versions of hapi and vision. 

Usage of those versions by customers is highly discouraged. Moving forward the Node APM Agent only supports the scoped versions of Hapi (20+) and plugins.

### Auto Instrumented packages:
[@hapi/hapi](https://www.npmjs.com/package/@hapi/hapi)
[@hapi/vision](https://www.npmjs.com/package/@hapi/vision)

### Unsupported deprecated packages:
[hapi](https://www.npmjs.com/package/hapi)
[vision](https://www.npmjs.com/package/vision)

### Notes:
- Pull Request merges into `nh-main`. Agent built from `master` will maintain support for old outdated `hapi`.
- Integration testing completed successfully.


test/probes/@hapi/hapi.test.js
```diff
@@ -329,7 +300,7 @@ describe(`probes.${hapiName} ${pkg.version} ${visionText}`, function () {
   it('should allow a custom TransactionName', function () {
     // supply a simple custom function
     function custom (request) {
-      const result = 'new-name.' + request.method + request.route.path
+      const result = 'hapi.hello.' + request.method + request.route.path
```
cheempz, line 303: The custom transaction SDK example was removed in https://github.com/appoptics/appoptics-apm-node/pull/197/files#diff-9eddf4dc51bf6a0125ca7fb094468ad284112270385c41cebce4f8f0a29620ab and this test case plus the one below for custom transaction with domain prefix also changed to what I _think_  is the unaltered/default transaction name that instrumentation would have set, i.e. no longer testing that we can set a different transaction name.   Is that the correct understanding, and if so, let's discuss what it takes to get custom transaction name working for @hapi.

cheempz (commented):
Left a comment @ronilan but the rest look good to my very untrained eyes :)

test/probes/@hapi/hapi.test.js
ronilan: Good Catch.

Custom transaction name **is working** for `@hapi/hapi`. API wasn't changed.

I put the example back as `hapi/hapi` (can't use @ sign there due to docs generator..) and fixed test so that it tests what it should be testing.

Side note: given the complexity of the tests, and scope of changes in `nh-main` it should be assumed that such "pass but not testing" tests are: a) already in the code (we've clean some in the past) and b) may be newly introduced (e.g. lambda). Review and simplification of tests is desirable at some point...

cheempz (approved):
LGTM, thanks @ronilan!

PR #200, opened 2021-12-02
NH-5315-Remove-Handle-Conflicts-Code

ronilan:
### Overview:

This pull request removes the code used to detect and potentially handle conflicts with `newrelic`, `strong-agent` and `appdynamics` and also removes the `ignoreConfilicts` config setting.

### Status:

The code, as implemented, checked the require cache to see if one of the ignored packages is loaded. However, since the AppOptics agent is  required, per our instructions, **before** any other packages, the agent will not be able to find anything in the required cache when the code is run.

Since this behavior is foundational, the code only checks edge-cases (AppOptics agent was not required first) of edge cases (other packages that may create conflicts were required prior). For simplification it has been removed.

### Notes:

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -30,7 +30,6 @@ The configuration file can supply the following properties, showing their defaul
   enabled: true,
   hostnameAlias: undefined,
   domainPrefix: false,
-  ignoreConflicts: false,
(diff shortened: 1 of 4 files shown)


cheempz (approved):
LGTM, thanks @ronilan!

PR #202, opened 2021-12-07
NH-5414-Update-Dependencies-devDependencies-1

ronilan:
### Overview:

This pull request removes `array-flatten` from dependencies.

`array-flatten` was added as dependency in https://github.com/appoptics/appoptics-apm-node/commit/9b2a6a058260350ed7091ea12673d0c352f8587c. 

### Status:

No longer needed. Array.prototype.flat() is available in [node versions 11+](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat#browser_compatibility).

### Notes:

### DIFF
--- a/lib/probes/express.js
+++ b/lib/probes/express.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const flatten = require('array-flatten')
 const shimmer = require('shimmer')
 const methods = require('methods')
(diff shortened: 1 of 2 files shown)


cheempz (approved):
LGTM, thanks @ronilan!

PR #203, opened 2021-12-07
NH-5407-Remove-AO_DEBUG_-Conditionals-from-Code

ronilan:
This Pull Request removes conditionals that use internal environment variables are in API code.

Pull Request merges into nh-main.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -224,25 +224,7 @@ function definePropertiesOn (ao) {
   // continuation-local-storage.
   //
   const contextName = 'ao-cls-context'
-  const clsOpts = { captureHooks: process.env.AO_DEBUG_HOOKS }
-  if (process.env.AO_DEBUG_CONTEXT) {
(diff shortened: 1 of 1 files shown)


cheempz (approved):
Agree we should minimize undocumented conditionals in production code.  Debugging context is hard though, so wonder if you'll have an alternative to help better see this data for debugging purposes?

ronilan:
> Agree we should minimize undocumented conditionals in production code. Debugging context is hard though, so wonder if you'll have an alternative to help better see this data for debugging purposes?

The code is fine. It was most likely helpful at the time. Might be helpful in the future. It just should never have shipped on the main branch. Generally, if debugging has to be done post distribution, it should be done by installing package from branch. Specifically, right now, it is one of two places that use `.toString(1)`, which calls bindings to format a string. So if untangling is to be archived, this has to go.

PR #207, opened 2021-12-15
NH-5727-Trigger-Trace-Updates

ronilan:
### Overview:

This pull request changes naming of headers and reported KV pairs for Trigger Trace requests.
`pd-keys` is now to `sw-keys` and it is reported as `SWKeys` rather than `PDKeys`.

### Notes:


jiwen624 (approved):
LGTM!

PR #209, opened 2022-01-07
NH-4534-Trace-Context-in-Logs

ronilan:
### Overview:

This pull request modifies trace context in logs to match the [Open Telemetry Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/overview.md#plain-text-formats).

1. Modified API used for logging:
  - Removed `getFormattedTraceId` method ([added in v6.1](https://github.com/appoptics/appoptics-apm-node/blob/master/CHANGELOG.md#v610)).
  - Modified `insertTraceIdsIntoLogs` to generate an object with top key of `sw` rather than `ao`
  - Modified `insertTraceIdsIntoLogs` to generate an object with three fields: `trace_id`, `span_id`, `trace_flags`.

2. Updated `morgan` probe to use updated API.
  - Replaced `getFormattedTraceId` with `insertTraceIdsIntoLogs`.
  - Removed a "flaky conditional" that under certain conditions would have omitted the inserted string when a custom function is used. Now string is inserted in all cases.
  - Changed auto token to prefix `sw` rather than `ao`.

3. Updated tests to use updated API and probes.

4. Updated documentation.

### Notes:
  - Removing `getFormattedTraceId` is a step in the process of removing the formatting functionality of the `Event.toString`. It is implemented in the bindings and results in unneeded calls to the bindings and a maintenance overhead.
  - Originally tests were checking that logged trace id matches expected trace id obtained by the test suite. Now tests also check that span id matches. For the `morgan` probe tests, there are variations as to where the event should be obtained from.
  - Pull Request merges into `nh-main`. Agent built from master will be unchanged.
  - Integration testing completed successfully.
  - "delayed" `fs` span is [showing](https://st-nighthawk.solarwinds.cloud/115724866230088704/traces/9013D5A00BA23EB26FF233891EC9CE05/8367BF19D8885706/details) as [it did](https://my-stg.appoptics.com/apm/225844/services/ao-node-test/traces/7DB906228F38DBE1A52E9A834512376FC6B9EDD3) with pre-pull request code. The loggers themselves though do not have own spans....

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -74,7 +74,6 @@ module.exports = function (appoptics) {
     reportInfo,
     sendMetric,
     sendMetrics,
-    getFormattedTraceId,
     insertLogObject,
 
     // lambda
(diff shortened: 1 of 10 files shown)


lib/api.js
```diff
@@ -1474,19 +1433,28 @@ function getFormattedTraceId (options = {}) {
...
+ *     more: 'there will be an added sw property'
  * }))
  */
 function insertLogObject (o = {}) {
   // if truthy and tracing insert it based on sample setting. otherwise if 'always'
   // then insert a trace ID regardless. No explicit check for 'traced' is required.
   let last
-  if (ao.cfg.insertTraceIdsIntoLogs && (last = ao.lastEvent)) {
+  if ((ao.cfg.insertTraceIdsIntoLogs || ao.cfg.insertTraceIdsIntoMorgan) && (last = ao.lastEvent)) {
     if (ao.cfg.insertTraceIdsIntoLogs !== 'sampledOnly' || last.event.getSampleFlag()) {
```
prettycoder, line 1444: should `|| ao.cfg.insertTraceIdsIntoMorgan !== 'sampledOnly'` be added to this condition as well?

prettycoder (commented):
one question

lib/api.js
ronilan: Sharp eyes...

The answer to the one question is of the "**yes, but no, then maybe, so I don't know**" category.

**Yes**, adding the condition would cover all the cases and make the code clear, **but no**, because that code would be redundant as the sampledOnly case is covered in the probe returning an empty string before calling the api method above, **then maybe** the probe code could be changed to not do that, or both the api code and probe code could be changed to no have a probe specific config conditional in the general API, or maybe the probe specific global config setting could be dropped all together, or maybe... **so I don't know**... 

Ask was to trace context in logs... We'll (maybe) deal with the API/Config vars later

prettycoder: Do you have test coverage in case the probe code changes?

ronilan: If the [probe conditional](https://github.com/appoptics/appoptics-apm-node/blob/cd6f461af51f9e62a0d40df909d1d60f02ef0d0f/lib/probes/morgan.js#L50) is removed the then a test from [this test array](https://github.com/appoptics/appoptics-apm-node/blob/cd6f461af51f9e62a0d40df909d1d60f02ef0d0f/test/probes/morgan.test.js#L244) will fail. But, note caveat re: morgan tests in PR description and the general complexity of code/test interfaces...

PR #213, opened 2022-01-19
NH-7328-Remove-Support-for-pg-Package-Below-7

ronilan:
### Overview:

This pull request removes support for pg package versions under 7, and from now on assumes `pg` package used by customer is 7 or above.

### Status:

There was a diversion at version 7.0.0 which is 4 years old. [Current version](https://www.npmjs.com/package/pg) is 8.7.1 from 6 moths back. Based on `testeachversion` [run](https://github.com/appoptics/appoptics-apm-node/blob/master/docs/supported-components.human) we only support 7.9.0 and above for node versions 10, 12, and 8.0.3 and above for node versions 14, 16, 17. Anything below can be considered "dead code".

### Notes:


test/probes/pg.test.js
```diff
@@ -1,17 +1,423 @@
...
+try {
+  require('pg/lib/native')
+  nativeVer = require('pg-native/package').version
+  hasNative = true
+} catch (e) {
+  ao.loggers.test.info('test/probes/pg 6+ failed to load pg native')
+  hasNative = false
 }
+
+describe(`probes.pg6+ ${pkg.version} pg-native ${nativeVer}`, function () {
```
prettycoder, line 42: nit: call it pg7+ ?

prettycoder (approved):
one comment

test/probes/pg.test.js
ronilan: better still - just call it `pg`. c1c53e7

PR #219, opened 2022-02-04
NH-5772-Trace-Context-in-Queries

ronilan:
## Overview 

This pull request adds injection of Trace Context into database queries. The functionality is meant to allow integration with [SolarWinds DPM](https://docs.vividcortex.com/) using [Query Tags](https://docs.vividcortex.com/general-reference/query-tags/). Tags used are similar to those inserted by [sqlcommenter](https://github.com/open-telemetry/opentelemetry-sqlcommenter).

## Implementation

Changes are contained within the `mysql`, `pg`, `oracledb` and `tedious` probes without breaking any API barrier.

### Configuration 

A new configuration value is introduced to each of the probes: `sqlTag`. It defaults to `false`.

### Tagging Decision

For all probes tagging is done when the probe is *registered* and *enabled* and `sqlTag` is set to `true`, we are *tracing* and we are *sampling*.

Additionally, for the `pg` probe only, named prepared statement are not tagged.

### Tag (SQL comment) Creation

A new module, `sql-trace-context` is introduced. It handles creation of an SQL comment string in the desired format.

Module exposes a single method `tag` which accepts a string and returns either a comment which contains a `traceparent` key and a single-quoted value that matches open telemetry [traceparent format](https://www.w3.org/TR/trace-context/), or, if input not valid, an empty string. 

Integration with probes is done via this method.

### Tagging

Comment injection is done by string concatenation at the runner function. The traceparent string is obtained from the api (a global method accessing key on global object), validated using the above module and prepended to the SQL statement provided in the arguments passed to the instrumented function.

### Key Value Pairs

A key `QueryTag` is added. It contains the tag (SQL comment) injected into the SQL. It is generated by a call to the callback function of the Span generating function. Note that `QueryTag` and the injected comment itself are generated by the same mechanism and using the same value but they do not refer to the same string.

## Tests

All [tests pass](https://github.com/appoptics/appoptics-apm-node/runs/5070747451?check_suite_focus=true) for Node versions 14 and 16, 17.

Integration testing is done using a [simple app](https://github.com/appoptics/node-instrumented) that performs CRUD operations on all four dbs.

Screen Shot 2022-02-04 at 10 01 59 AM

Injection verification testing via query logging is *not* implemented at this time.

## Risk

The instrumentation in the package is based on the [shimmer package](https://www.npmjs.com/package/shimmer) which notes that: *"There are times when it's necessary to monkeypatch default behavior in JavaScript and Node. However, changing the behavior of the runtime on the fly is rarely a good idea ... "*

Given the above, the general approach to instrumentation was to *observe only*. The patched function collects the needed tracing data and then apply the original function to the user supplied arguments *as-provided*.

This feature is a departure from that concept. With this pull request merged, the agent will be **modifying user supplied arguments for mission critical database interactions**.

### Notes:

### DIFF
--- a/lib/probe-defaults.js
+++ b/lib/probe-defaults.js
@@ -11,9 +11,11 @@ module.exports = {
   // NOTE: Disabling backtraces may improve performance, but will make data
   // presented in the dashboard less useful for correlating code locations.
   //
-  // SQL database modules also include a sanitizeSql setting to prevent
-  // potentially sensitive data from being reported to AppOptics.
-  //
+  // SQL database modules also include:
+  //  - a sanitizeSql setting to prevent potentially sensitive data from being
+  //    reported to AppOptics (default: true).
+  //  - a tagSql setting to inject a traceparent value a comment into SQL
+  //    statements sent to database (default: false).
 
   //
   // node-native modules
@@ -78,7 +80,8 @@ module.exports = {
   'cassandra-driver': {
     enabled: true,
     collectBacktraces: true,
-    sanitizeSql: true
+    sanitizeSql: true,
+    tagSql: false
   },
 
   // https://npmjs.org/package/co-render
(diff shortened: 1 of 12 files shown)


jerrytfleung (approved):
LGTM

ronilan:
Feature does not LGTM. Merging based on organizational decision.

PR #227, opened 2022-02-28
NH-9004-Update-Support-for-Pino-6.x-7.x

ronilan:
## Overview

This pull request updates the pino probe so that it supports instrumenting all latest versions.

## Status

[Pino](https://www.npmjs.com/package/pino) is at Version 7.8.0. Probe only support up to 5.17.0 Both 7.x and 6.x are actively maintained. 5.x is EOL at 5.17.0 (see: https://www.npmjs.com/package/pino?activeTab=versions)

In [version 6.4.0](https://github.com/pinojs/pino/releases/tag/v6.4.0) pino introduced instantiation via factory (see: https://github.com/pinojs/pino/pull/875) as a result the patching method used by the probe (on the prototype) does not work anymore.

## Implementation

The result of requiring pino is an anonymous function from which logger instance objects are created (`const logger = require('pino')()` see: https://github.com/pinojs/pino#usage).

The implementation in this PR uses a "nested wrapper". It patches the pino anonymous function so that when it is invoked, it will patch an internal write function of the returned logger object. That write function is where trace context is added to the logs. 

The initial wrapper also augments the patched anonymous function with the properties of the original anonymous function so that the patched pino object resembles the original.

## Notes:
- Version has been removed in [version 6.0.](https://github.com/pinojs/pino/releases/tag/v6.0.0) of pino. Hence test were modified. (see: https://github.com/pinojs/pino/pull/623).


test/probes/pino.test.js
```diff
@@ -117,51 +90,25 @@ describe(`pino v${version}`, function () {
...
-            console.log(s);  // eslint-disable-line
-          }
-        },
-        flush () {},
-        flushSync () {}
-      }
-      Object.setPrototypeOf(stream, EventEmitter.prototype)
+      logger[pino.symbols.streamSym] = modStream
+      Object.setPrototypeOf(logger[pino.symbols.streamSym], EventEmitter.prototype)
     } else if (major >= 2) {
```
prettycoder, line 109: still needed? isn't it all >= 5 now?

package.json
```diff
@@ -108,7 +108,7 @@
     "oracledb": "^5.2.0",
     "pg": "^8.7.1",
     "pg-native": "^3.0.0",
-    "pino": "^5.17.0",
+    "pino": "^7.8.0",
```
prettycoder, line 111: big jump, I assume you tested with 6 as well

prettycoder (approved):
looks good, some comments

package.json
ronilan: Yep. All good.

test/probes/pino.test.js
ronilan: Thought about removing it - but decided to keep it as it works. We'll need to make a decision regarding "floor" of versions across all probes and clean there. Someday.

ronilan: Versions before 5 are still instrumented.

I thought about removing them - but since they work - I'll keep it for now.
We need to make a decision regarding "floor" versions across all probes and how to handle it. Someday.

PR #228, opened 2022-03-10
NH-5459-Support-Trace-ID-Insertion-for-log4js

ronilan:
## Overview 

This pull request adds auto instrumentation of the [`log4js`](https://www.npmjs.com/package/log4js) logging package to the agent.

## Implementation

A new probe named `log4js` has been added.

In log4js logging is done by calling a level function (e.g `logger.debug('say something')`) which then calls the `log` function, or alternatively, by directly calling the `log` function. Insertion is done by patching the lower level `log` function to cover all cases (including custom levels).

The implementation uses a "nested wrapper". It patches the `getLogger`  function so that when it is invoked, it will patch an internal `log` function of the returned logger object.

In order to provide granular control as what layouts are auto patched, and which is left for the user to configure, The implementation patches the `configure` function so that it saves the applied configuration on the object so that it can in turn be read at instantiation time.

### Usage
When enabled and inserting (see configuration below) the probe will append a space delimited trace data string to the end of pre predefined layout types. For example, using the `basic` layout: 
`[2022-03-09T20:44:54.716] [ERROR] default - Cheese is too ripe!` will become:
`[2022-03-09T20:44:54.716] [ERROR] default - Cheese is too ripe! trace_id=d91f00e5f3cad1db5f13814aa09cfcd2 span_id=e9f3d82c1dbfb401 trace_flags=01`.

When the user is using a pattern in their layout in ANY of their appenders the probe will respect their pattern (and skill)
and will **NOT** append the trace to the end of the log line. 

Adding trace data to patterns is supported using log4js tokens and custom layouts and the newly added API method `getLogString` (see note re: naming below).

### Configuration 

Probe has basic configuration values that are set in a manner similar to all other logging probes.

 ```enabled: true```

 It obeys the global config ```insertTraceIdsIntoLogs: false``` see: [Configuration Guide](https://github.com/appoptics/appoptics-apm-node/blob/NH-5459-Support-Trace-ID-Insertion-for-log4js/CONFIGURATION.md)

### Tests

Tests added for probe. Tests include testing case where `log4js` is used as `express` middleware (using built-in `connectLogger`).

All [tests pass](https://github.com/appoptics/appoptics-apm-node/actions/runs/1961201242).

Integration testing is done using a [simple app](https://github.com/appoptics/node-instrumented/tree/main/log4js) that performs logging operations with multiple configurations. Those include all predefined configurations, configuration with multiple appenders, configuration using tokens, configuration with custome log levels as well as using built-in addContext and addLayout methods.

### Notes:
- Pull request is a first in three to revamp logging probes. Final name for API method will be `getTraceStringForLog` and will be applied in a subsequent pull request.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -75,6 +75,7 @@ module.exports = function (appoptics) {
     sendMetric,
     sendMetrics,
     insertLogObject,
+    getLogString,
 
     // lambda
     wrapLambdaHandler // wrap user handler
@@ -1405,7 +1406,7 @@ function sendMetrics (metrics, gopts = {}) {
  * @method ao.insertLogObject
  * @param {object} [object] - into which an sw log object containing trace_id, span_id, trace_flags
  * properties is inserted when conditions are met.
- * 
+ *
  * @returns {object} - the object with the an additional properties, ao, e.g.
  * object.sw === {trace_id:..., span_id: ..., trace_flages: ...}.
  *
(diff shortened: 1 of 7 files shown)


lib/api.js
```diff
@@ -1459,6 +1460,44 @@ function insertLogObject (o = {}) {
...
+ *         }
+ *       }
+ *     }
+ *   },
+ *   categories: { default: { appenders: ['out'], level: 'info' } }
+ * })
+ * const logger = log4js.getLogger()
+ * logger.info('token from api')
+ */
+function getLogString (delimiter = ' ') {
```
prettycoder, line 1496: should delimiter be checked so we don't expose a log injection vulnerability?

prettycoder (approved):
one question, you decide

lib/api.js
ronilan: Interesting... didn't think about that... but since it is directly passed to a native `'.join()` which in turn casts whatever it gets into a string (or if it doesn't, than this is not "our vulnerability" ), I'll leave it as is.

PR #230, opened 2022-03-10
NH-9492-Refactor-Morgan-Probe

ronilan:
## Overview

This pull request refactors the `morgan` probe.

## Issues

- Probe is implemented in a unique manner using a Proxy. It does not use the [shimmer third party package](https://www.npmjs.com/package/shimmer) used by all other probes. It’s a nice concept, but it “breaks the pattern”.
- Probe uses a probe specific config `createTraceIdsToken` that is conditionally nested under (now deprecated) `insertTraceIdsIntoMorgan`, a pattern none of the other probes use. The need for that specific setting is also unclear as creating a morgan token is part of the "normal flow" of instrumentation.
- Trace data insertion into predefined format is done with a token, but is then unnecessarily conditionally checked each time logging is invoked.
- Trace insertion into the predefined dev format (which is a compiled function) does not actually work.
- Does not have test coverage for all config options (cases), hence test pass, but should not.
- Tests test for things that are test specific and those may change when all config options are tested.

## Refactor Implementation

Implementation greatly simplifies how the probe works.

When the morgan probe is enabled, at patch time, and regardless of `insertTraceIdsIntoLogs` setting, a morgan token named :`trace` is created. It is used for all insertions, both automatic and and user defined (see below).

Because the dev predefined format is a compiled function, there is no access to the format string. To archive insertion the function is patched.

User defined format strings and functions are untouched.

### Usage

When enabled and inserting the probe will append a space delimited trace data string to the end of pre predefined layout types. 
For example, using the `combined` layout: 
- `::ffff:172.27.0.1 - - [10/Mar/2022:19:39:46 +0000] "GET / HTTP/1.1" 200 - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.51 Safari/537.36"` 

will become:
- `::ffff:172.27.0.1 - - [10/Mar/2022:19:39:46 +0000] "GET / HTTP/1.1" 200 - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.51 Safari/537.36" trace_id=1331bc8c4d2eb26343234948e2ce3ac8 span_id=3588288099552501 trace_flags=01`.

When the user is using a string pattern the probe will respect their pattern (and skill) and will **NOT** append the trace to the end of the log line. User can use the `:trace` token in custom format strings (e.g. `:method :url :trace`) to add the trace data, as well as in custom functions (e.g. `tokens.trace()`).

### Configuration 

Probe has basic configuration values that are set in a manner **similar to all other logging probes**.

 ```enabled: true```

 It obeys the global config ```insertTraceIdsIntoLogs: false``` see: [Configuration Guide](https://github.com/appoptics/appoptics-apm-node/blob/NH-5459-Support-Trace-ID-Insertion-for-log4js/CONFIGURATION.md)

### Tests

Test coverage was significantly increased and test setup has been (mildly) simplified.

All [tests pass](https://github.com/appoptics/appoptics-apm-node/runs/5514511552?check_suite_focus=true).

Integration testing is done using a [simple app](https://github.com/appoptics/node-instrumented/tree/main/morgan) that performs logging operations with multiple configurations. Those include all predefined configurations (including precompiled `dev`), configuration using a custom string pattern (with and without token), and configuration using a custom function (with and without using the token method).

### Notes:

### DIFF
--- a/lib/probes/morgan.js
+++ b/lib/probes/morgan.js
@@ -12,7 +12,7 @@ module.exports = function (morgan, info) {
 
   // insertion into predefined formats is done by adding a token to the predefined format string
   // user can also use token in their own formats
-  const autoToken = ':sw-auto-trace-id'
+  const autoToken = ':trace'
 
   // morgan tokens are functions
   // define a token function at load time
--- a/lib/probes/morgan.js
+++ b/lib/probes/morgan.js
@@ -1,135 +1,42 @@
 'use strict'
 
 const ao = require('..')
+const shimmer = require('shimmer')
 
 const logMissing = ao.makeLogMissing('morgan')
 
-let token
-
-//
-// patch morgan so the arguments can be mucked with. this is a proxy
-// because 1) morgan is typically only called once or maybe twice and
-// 2) it will let us intercept token and format calls at some point if
-// we choose to do so.
-//
-module.exports = function (morgan) {
+module.exports = function (morgan, info) {
   if (!ao.probes.morgan.enabled) {
     return morgan
   }
 
-  const pmorgan = new Proxy(morgan, {
-    // call target/morgan as a function
-    apply (target, thisArg, argumentsList) {
-      if (ao.cfg.insertTraceIdsIntoLogs) {
-        argumentsList = tryInsertion(morgan, argumentsList)
-      } else if (ao.cfg.createTraceIdsToken === 'morgan') {
-        if (!token) {
-          token = createToken(morgan)
-        }
-      }
-      return Reflect.apply(morgan, thisArg, argumentsList)
-    }
+  // insertion into predefined formats is done by adding a token to the predefined format string
+  // user can also use token in their own formats
+  const autoToken = ':sw-auto-trace-id'
+
+  // morgan tokens are functions
+  // define a token function at load time
+  morgan.token(autoToken.slice(1), () => ao.getTraceStringForLog())
+
+  // do format modification
+  const predefineds = ['combined', 'common', 'short', 'tiny']
+  predefineds.forEach(predefined => {
+    morgan.format(predefined, `${morgan[predefined]} ${autoToken}`)
   })
 
-  return pmorgan
-}
+  if (typeof morgan.dev === 'function') {
+    // the dev predefined format is a compiled function. thus there is no access to the format string
+    // wrap it to modify function result
+    shimmer.wrap(morgan, 'dev', function (original) {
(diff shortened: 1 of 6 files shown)


test/probes/morgan.test.js
```diff
@@ -323,6 +349,127 @@ describe(`probes.morgan ${version}`, function () {
...
+        function (msg) {
+          msg.should.have.property('Layer', spanName)
+          msg.should.have.property('Label', 'exit')
+        }
+      ], localDone)
+    })
+  })
+
+  //
+  // for each mode verify that insert works with format function
```
cheempz, line 435: To confirm, the test case below verifies if a custom function is used to define the log message but our trace context is not requested (via using the `tokens` property) we don't insert.

Can we add a test case where the custom function uses the `tokens.trace()` to verify correct insertion if the trace context is requested?

lib/probes/morgan.js
```diff
@@ -1,135 +1,42 @@
...
-let token
-
-//
-// patch morgan so the arguments can be mucked with. this is a proxy
-// because 1) morgan is typically only called once or maybe twice and
-// 2) it will let us intercept token and format calls at some point if
-// we choose to do so.
-//
-module.exports = function (morgan) {
+module.exports = function (morgan, info) {
```
cheempz, line 8: Random curiosity, what is defining / providing `info` and how is it used?

cheempz (approved):
LGTM, thanks @ronilan!  I left a couple minor comments.

test/probes/morgan.test.js
ronilan: Indeed. Will do.

lib/probes/morgan.js
ronilan: It’s the signature of the require patch. Info is an object containing name and version (https://github.com/appoptics/appoptics-apm-node/blob/0b0591e61c598bc33e4bb27050012505b546e652/lib/require-patch.js#L98)

it’s used when probe needs variation in relation to original package version (e.g. https://github.com/appoptics/appoptics-apm-node/blob/0b0591e61c598bc33e4bb27050012505b546e652/lib/probes/pino.js#L25)

cheempz: I see, thanks :)

PR #241, opened 2022-04-18
NH-11221-Remove-bin-Functionality

ronilan:
## Overview

This pull request removes the `bin` functionality from the agent.

## Status

Sometime in the distant past an option was provided to run the agent via a command line argument. The feature is not documented and similar functionality is available via the `-r` option of node (e.g. node -r solarwinds-apm app.js`).

Removing the functionality will allow removing `minimist` from the dependency list.

## Notes:
- Pull Request merges into `nh-main` branch. Agent built from `master` will stay unchanged.

PR #243, opened 2022-04-18
NH-5414-Update-Dependencies-devDependencies-3

ronilan:
### Overview:

This pull request removes `minimist` from dependencies and moving it to devDependencies (where it is only used for utility by the, currently gracefully disabled, lambda implementation).

`minimist` was added as dependency in https://github.com/appoptics/appoptics-apm-node/commit/ce5890d2ab0e8be98b7ebae047c32f3c3de74e2f

### Status:

Need for package was removed in #241

### Notes:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "cls-hooked": "^4.2.2",
         "debug-custom": "^1.1.0",
         "glob": "^7.2.0",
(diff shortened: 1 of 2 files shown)


cheempz (approved):
LGTM, thanks @ronilan!

PR #153, opened 2021-08-19
AO-20334-Create-GitHub-Actions-Test-Publish-Workflows-3

ronilan:
This Pull Request is the **3rd** in a series of Pull Requests to set up GitHub Actions.

### Commits 
Commit (859ab3c) adds a Document workflow triggered manually, supporting scripts in `.github/scripts` and a target directory (`docs`) that is currently empty.

Workflow runs a [`testeachversion`](https://www.npmjs.com/package/testeachversion) per package in a matrix, generates a summary file, creates an automated commit into branch and opens a pull request. End result is a docs directory containing supported-components.human file and supported-components-data directory with summary and raw data for each package.

### General Notes:

- The workflow is of the "dip a toe in the water" type. If all works as expected over time, it will replace https://github.com/appoptics/apm-test-node and the corresponding @aws setup.

- `testeachversion` is used as-is. Since it was not built for parallel run on @github @actions, some “rough patching” is required.

- Main goal is to provide additional confidence en route to end-to-end package publication using GitHub Actions (this repo and upstream repo).

### Result Notes:
- Since this is a manually triggered workflow that is on a branch, we can not run it before we merge it to master. There is however a sample at the dev repo:
    - Workflow run: https://github.com/appoptics/appoptics-apm-node-dev/actions/runs/1148854623
    - Pull Request: https://github.com/appoptics/appoptics-apm-node-dev/pull/89
    - Downloadable Results: https://github.com/appoptics/appoptics-apm-node-dev/suites/3547083183/artifacts/84988458
 
- Workflow uses `continue-on-error: true` so that if a set of tests fails for a specific version of a package, the whole workflow will still complete. Thus in the run above the jobs are all ✅ while there is annotations below suggesting failures ❌. 

- Those annotated failures are as-should-be and the failing package version is removed from supported components list. To illustrate:
   - Annotation is: ❌ **Test Each Version (10, amqplib)** Process completed with exit code 1.
[...]

PR #257, opened 2022-04-29
AO-21197-Update-Document-Workflow-to-only-Commit-Summary

ronilan:
## Overview

This pull request simplifies the data generated by the Document Workflow, by trimming the data it generates. The 

## Status

The Document Workflow was implemented as an "experimental" feature with a goal is to provide confidence en route to end-to-end package publication using GitHub Actions (this repo and upstream repo). (see: #153). Now that end-to-end publication is working (for four packages...) it is time to simplify.

## Change

Workflow runs a [`testeachversion`](https://www.npmjs.com/package/testeachversion) per package in a matrix, generates a summary file, creates an automated commit into branch and opens a pull request. End result now that only `supported-components.human` file is placed in the docs directory.

The `supported-components-data` directory with summary and raw data for each package is no longer committed. It is still available as a workflow artifact.

## Test

- #256 was generated by the modified workflow against own branch (`AO-21197-Update-Document-Workflow-to-only-Commit-Summary`), merged into it and is now part of this pull request.

- The simplified diff presented by #256 is here: https://github.com/appoptics/appoptics-apm-node/pull/256/files


cheempz (approved):
LGTM, thanks @ronilan!

PR #258, opened 2022-04-30
NH-5414-Update-Dependencies-devDependencies-4 (glob)

ronilan:
## Overview

This pull request removes `glob` from dependencies.

`glob` was added as dependency in https://github.com/appoptics/appoptics-apm-node/commit/52b2728bfffc86ac2bb03ae67268f4f0846af8c6

## Status:

Package is used once in `require-patch` to dynamically load [a list of probes from a list of files](https://github.com/appoptics/appoptics-apm-node/blob/master/lib/require-patch.js#L126). Since the list of probes is a static attribute of the package, doing so dynamically is an anti-pattern (which has been used from the beginning, even prior to introducing the dependency).

## Change

Instead of dynamic load, list of probes is loaded from probe config file. As a (desired) side-effect, listing a probe in the config file is now mandatory.

## Notes:


lib/require-patch.js
```diff
@@ -1,22 +1,13 @@
...
 const probes = {}
 
-// allow scoped packages to use same tests as unscoped versions. this
-// facilitates testing without duplicating for supported packages that
-// change to scoped.
-const nameToPatchFile = {
-  hapi: '@hapi/hapi',
-  vision: '@hapi/vision'
-}
-
```
cheempz, line 19: Does this impact whether the probe tests run for hapi?  Quick search I'm not seeing any test cases for either `hapi` or `@hapi` from the GH "Full Test" workflows... could be looking in the wrong place but want to double check.

cheempz (commented):
Thanks for the detailed PR writeup and would be great to remove another unneeded dependency @ronilan!  I did have a few comments, and a more general question:

The `probe-defaults.js` file might have started life as something customers can directly update to configure the agent.  Is it critical now that the customer should not touch it?  If so, we should update the comments in that file and the https://github.com/appoptics/appoptics-apm-node/blob/NH-5414-Update-Dependencies-devDependencies-4/CONFIGURATION.md#configuration-file-probe-settings wording to reflect.

lib/require-patch.js
ronilan: No. But...
The tests `hapi` and `vision` tests were indeed not running as they were not picked by the [test runner](https://github.com/appoptics/appoptics-apm-node/blob/d91896364bcc2a95e55321b471c7c594e3d74324/test.sh). That was "lost" in #197 (or followup rebases).  Fixed with 489392c6946961fbaddc3655d77cb32b15e6ce0b. See [Full Test Run](https://github.com/appoptics/appoptics-apm-node/runs/6284568075?check_suite_focus=true).

lib/require-patch.js
```diff
@@ -29,11 +20,6 @@ exports = module.exports = {
   //
   register (name, path) {
     probes[name] = path
```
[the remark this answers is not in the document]
ronilan: No more that any other function call can have "bad input". `name` is the probe name in the config file. Previously it was the file name in the directory. Both created by developer.

The code removed configured a probe not in the configuration file. That case is no longer possible.

ronilan:
The `probe-defaults.js` file is stashed deep in `node_modules` - so nothing the customer ever touches. The [config file](https://github.com/appoptics/appoptics-apm-node/blob/nh-main/appoptics-apm-config.json) however contains a `probs` object that may include anything in the probe defaults file AND there is no documentation of those options other than what is in `probe-defaults.js`. It's a known issue (See: https://swicloud.atlassian.net/browse/AO-17192) I'll pull it out of the backlog and generate some `.md` to make things clearer.

ronilan:
After taking another look at the code it was further simplified. see: e7edd63494169dee0dc1a409e3e889538a0b7e5a

lib/require-patch.js
cheempz: 👍

lib/require-patch.js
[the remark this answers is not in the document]
cheempz: 👍

cheempz (approved):
LGTM, thanks for the revisit @ronilan!

PR #259, opened 2022-05-20
AO-21254-Change-Release-Workflow-to-Manual-Trigger

ronilan:
## Overview

This pull request changes the release workflow so that it is triggered manually.

## Status

Current release workflow is triggered by tag push (that was generated with `npm version`). This is fragile under the "two repo" regime (appoptics/solarwinds) and requires more developer involvement.

## Change

* Workflow is triggered manually.
* User is required to type the version being released as it is listed package.json to avoid "accidental clicking".
* Before running workflow will check that:
  - User input of version is valid and matches what is in `package.json`
  - Version being released is a full semver or has a `-prerelease.` preid
  - Version being released has never been published.

## Notes
- Since this is a manually triggered workflow pushed for the first time, it is NOT possible to trigger it on the branch before it is merged.
- Similar setup used by bindings worked as expected.
- Testing on dev repo (excluding last NPM publish step) works as expected.
- Prior to the change, using `npm version` which modifies `package.json`, and creates a git tag and commit, was required for release. It is now optional to the developer. If modifying `pacakge.json` by hand, care is due.


.github/workflows/release.yml
```diff
@@ -1,21 +1,33 @@
...
+    steps:
+      - name: Checkout ${{ github.ref }}
+        uses: actions/checkout@v2
+
+      - name: Setup Node 14
+        uses: actions/setup-node@v2
+        with:
+          node-version: '14.x'
+
+      - name: Check is pablishable version
```
cheempz, line 24: Minor publishable typo

cheempz (approved):
LGTM @ronilan, thanks for the docs update too.

cheempz (approved):
👍

PR #267, opened 2022-06-02
NH-14130-Fix-http-Tests

ronilan:
## Overview
This pull request fixes the http probe tests so that they actually run as expected.

## Status

Since the tests did not run as expected, an issue with bindings/liboboe trace continuation when Downstream was not detected.

## Change

With the fixed tests, a test will fail when using bindings below version `12.0.0-nh.2` (comes with liboboe `10.5.0`).

See: https://github.com/appoptics/appoptics-apm-node/runs/6713249942?check_suite_focus=true
```
57 suites in 5 groups passed
2 suites in 1 group failed
    PROBES:
    test/probes/http.test.js
    test/probes/https.test.js
```
All tests will pass once https://github.com/appoptics/appoptics-bindings-node/pull/118 is merged and the updated bindings are installed in the agent `nh-main` branch from the bindings `nh-main` branch. see: https://github.com/appoptics/appoptics-apm-node/actions/runs/2429881575

## Notes
- Pull Request merges into nh-main branch. Bindings built from master will stay unchanged.
- Binding change will arrive at SolarWinds APM in next release.

### DIFF
--- a/test/probes/http-common.js
+++ b/test/probes/http-common.js
@@ -180,18 +180,19 @@ describe(`probes.${p}`, function () {
           expect(msg).property('Status', 200)
         }
       ], function () {
-        server.close(done)
+        server.close()
       })
 
       server.listen(function () {
         port = server.address().port
         axios(
-          `${p}://localhost:${port}/foo?bar=baz`,
-          function (error, response, body) {
-            expect(response.headers).exist()
-            expect(response.headers).property('x-trace')
-          }
-        )
+          `${p}://localhost:${port}/foo?bar=baz`
+        ).then(function (response) {
+          expect(response.headers).property('x-trace')
+          done()
+        }).catch(function (err) {
+          done(err)
+        })
       })
     })
 
@@ -204,20 +205,16 @@ describe(`probes.${p}`, function () {
         res.end('done')
       })
 
-      let trasportXtrace = ''
-
       helper.doChecks(emitter, [
         function (msg) {
           check.server.entry(msg)
           expect(msg).not.have.property('Edge')
-
-          trasportXtrace = msg['X-Trace']
         },
         function (msg) {
           check.server.exit(msg)
         }
       ], function () {
-        server.close(done)
+        server.close()
       })
 
       server.listen(function () {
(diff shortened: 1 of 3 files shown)


test/probes/http-common.js
```diff
@@ -313,22 +305,18 @@ describe(`probes.${p}`, function () {
...
 
-      let trasportXtrace = ''
-
       helper.doChecks(emitter, [
         function (msg) {
           // the edge in "Continuation" is from trace parent.
           expect(msg).property('Edge', otherTracestateOrgPart.slice(3).split('-')[0].toUpperCase())
           expect(msg).property('sw.tracestate_parent_id')
           expect(msg).property('sw.w3c.tracestate')
-
```
cheempz, line 324: The comment in line 310 says that edge should be from the traceparent, which is correct.  But the assertion after that seems to check the edge is from tracestate instead.  I'd exepct the `sw.tracestate_parent_id` property to capture our upstream span ID carried in tracestate, and also that `sw.w3c.tracestate` is only set if there was other vendor information in the tracestate header.

cheempz (commented):
Looks good pending one question about the "continuation" test scenario.

test/probes/http-common.js
ronilan: The comment was wrong. 

Fixed.

Note:
* A *Flow* request is one that is directly downstream from a SolarWinds APM instrumented app. It has valid, `traceparent` and `tracestate` headers. Org <u>is in</u> `tracestate`. The span id part of `traceparent` <u>match</u> the span id part in `tracestate`. 
* A Continuation request is one that comes directly downstream form an Open Telemetry instrumented service, but has passed through a SolarWinds APM instrumented service in the past. It has valid `traceparent` and `tracestate` headers. Org <u>is in</u> `tracestate`. The span id part of `traceparent` <u>does not match</u> the span id part in `tracestate`.

The Edge always goes back to "our" span (if exist). It this case (Continuation) it goes back to the one before the OTEL chasm, which is stored in tracestate, thus creating the link over the chasm.

In this test headers are:
```
  traceparent: '00-11111111111111111111111111111111-7a71b110e5e3588d-01',
  tracestate: 'sw=9999888855667788-01',
```
The span id part of `traceparent` <u>does not match</u> the span id part in `tracestate` so the edge to "our" span is from tracestate.

ronilan: See edge defined in: https://my.dc-01.dev-ssp.solarwinds.com/138094000386806784/traces/5287C1AE1729E00EC45639067B9ECB0A/421F401A9A334CE7/graph#event-98E5B0ECE32AD8B5 for SolarWinds APM (S) linking over 3rd party OTEL chasm (X).

ronilan: The current implementation ins the correct implementation.

ronilan: See "standard OTel-generated segments" in: https://my.dc-01.dev-ssp.solarwinds.com/138094000386806784/traces/5005BCD29749B037E83ECC55F6CE9CC0/5CE3ABDA691741D9/details for SolarWinds APM (S) linking to vanilla OTEL based JavaScript Agent (O). (chain SOSSOSSOS)

Screen Shot 2022-06-09 at 1 40 48 PM

cheempz (approved):
As discussed we'll address the trace context propagation in a separate PR, approving to unblock the ability to run tests correctly.

PR #266, opened 2022-06-01
NH-14040-Replace-websocket.org-in-Tests

ronilan:
## Overview

This pull request "unskips" web socket tests that were skipped in https://github.com/appoptics/appoptics-apm-node/commit/7c3230da467dac81640a2f1a0889c10fec086a57 replacing the now dead websocket.org with websocket.events.

websocket.events is operated by Lob and they say ["You’re also welcome to use ours at echo.websocket.events."](https://www.lob.com/blog/websocket-org-is-down-here-is-an-alternative) - so we'll do for now.

## Notes
- Pull Request merges into nh-main branch.

PR #272, opened 2022-06-16
Revert "NH-14040-Replace-websocket.org-in-Tests"

ronilan:
Reverts appoptics/appoptics-apm-node#266

Tests fail randomly too many times while running in GitHub Actions.


cheempz (approved):
LGTM but curious, are the tests failing due to public server flakiness or something brittle about the tests themselves?

ronilan:
> LGTM but curious, are the tests failing due to public server flakiness or something brittle about the tests themselves?

Good question. Public server is flaky, but, I think there may be issues with the instrumentation itself and thus with how it is tested.... Needs more carful attention.

PR #273, opened 2022-06-17
NH-5414-Update-Dependencies-devDependencies-6 (debug-custom)

ronilan:
## Overview

This pull request "unbundles" `debug-custom` package dependency into local module. That local module depends on the actively maintained `debug` package.

## Status:

[`debug-custom`](https://www.npmjs.com/package/debug-custom) was introduced in https://github.com/appoptics/appoptics-apm-node/pull/54 replacing [`debug`](https://www.npmjs.com/package/debug)

It was made by former maintainer for use in the Agent. 
Readme says: 
> “... debug-custom is a thin wrapper around `debug` that allows the user to hide whatever prefix is used to prevent conflicts. debug-custom also makes it easier to add and remove debug levels dynamically under program control.” 

## Discussion

- As of June 17, 2022
   - `debug` Last published 3 months ago ago, has 182,971,812 Weekly Downloads and 43,521 Dependents.
   - `debug-custom` was Last published 2 years ago, has 3,037 Weekly Downloads and 3 Dependents (2 of which are SolarWinds Agents).

- Ideally, if a dependency is required, it is better to use one that is well maintained, popular *and* substantial (i.e. not a "thin wrapper"). 

- Since `debug-custom` is written in plain JavaScript (no transpiling and packaging involved) and is [MIT licensed](https://github.com/bmacnaughton/debug-custom/blob/master/LICENSE) "unbundeling" is possible and simple.

- "unbundeling" allows:
  - Keeping functionality as-is.
  - Ability to keep all dependencies up to date.
  - Control of the code in case change is needed due to upstream breaking changes (security).

## Change

- Code from https://github.com/bmacnaughton/debug-custom/blob/master/lib/index.js was copied verbatim to a local module file.
- Local module was required instead of package.
- `debug-custom` was removed from dependencies. `debug` was added to dependencies.

## Notes:

### DIFF
--- a/lib/debug-custom.js
+++ b/lib/debug-custom.js
@@ -1,0 +1,144 @@
+/**
+ * A simple class built on top of the debug package that enables using a custom
+ * environment variable and avoids having to specify prefixes for each log level.
+ */
+
+class Logger {
+  constructor (name, opts = {}) {
+    this.name = name // the name of the environment variable storing the settings
+    this.prefix = name + ':' // the prefix used when setting debug's DEBUG env var
+    let internal = [] // settings prefixed with `${name}:`
+    const external = [] // settings other than `${name}:` prefixed
+    this.current = [] // an array of non-prefixed current settings for name
+    const envName = opts.envName || name
+
+    const env = process.env
+
+    this.originalExternal = env.DEBUG
+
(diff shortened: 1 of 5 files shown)


cheempz (approved):
LGTM

PR #275, opened 2022-06-20
NH-13380-Resolve-Circular-Dependencies

ronilan:
## Overview

This pull request resolves circular dependencies in agent code.

## Status

- `/log-formatters.js` was added in https://github.com/appoptics/appoptics-apm-node/commit/8b8e19b38fbeb28cd618edf0ad5f0e7dc7716b32 and https://github.com/appoptics/appoptics-apm-node/commit/e6cff9cb0a959235e2598361e313129cc00ced16 created a circular dependency (2018).

- Running [Madge](https://www.npmjs.com/package/madge):

```
Processed 58 files (1.3s) (1 warning)

✖ Found 2 circular dependencies!

1) index.js > loggers.js > log-formatters.js
2) index.js > require-patch.js
```

## Discussion:

Node documentation says: 
> Careful planning is required to allow cyclic module dependencies to work correctly within an application. https://nodejs.org/api/modules.html#modules_cycles. 

However at this point we can assume this is not a cause of customer issues.

## Change

- [ee2f556](https://github.com/appoptics/appoptics-apm-node/pull/275/commits/ee2f55639bbbd74135397e39908896519eba1f15) Resolved circular dependencies in `require-patch` by directly requiring the specific modules needed (loggers, probe-defaults)
- [5282aff](https://github.com/appoptics/appoptics-apm-node/pull/275/commits/5282affe608eb76b73df96b7ca974d44d9585c89) Resolved circular dependencies by removing reference to ao object. Instead of compering and object `instanceof` in conditionals, now using the self evident properties of the evaluated object.

- Running [Madge](https://www.npmjs.com/package/madge):

```
Processed 58 files (1.4s) (1 warning)

✔ No circular dependency found!
```

## Notes:
- Pull Request merges into `nh-main`. Agent built from `master` will stay unchanged`.

### DIFF
--- a/lib/log-formatters.js
+++ b/lib/log-formatters.js
@@ -1,12 +1,12 @@
 'use strict'
+
+const format = require('util').format
+
 /**
  * Define AppOptics-specific formatters for debugging and logging.
  */
 
 module.exports = function (d) {
-  const ao = require('./')
-  const format = require('util').format
-
   /**
    * format like console.log when replacing stdout, stderr
    */
@@ -50,11 +50,10 @@ module.exports = function (d) {
       return '<undefined>'
     }
 
-    if (event instanceof ao.Event) {
-      return event.Layer + ':' + event.Label + ' ' + humanID(event.event)
-    }
+    const layer = event.Layer || '?'
+    const label = event.Label || '?'
 
-    return '?:? ' + humanID(event)
+    return `${layer}:${label} ` + humanID(event)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lib/log-formatters.js
```diff
@@ -80,19 +79,20 @@ module.exports = function (d) {
    * Format an X-Trace ID in easier to look at format.
    */
   function humanID (x) {
-    if (x instanceof ao.addon.Event) {
+    if (x.constructor.name === 'Event' && typeof x.event === 'undefined') {
```
cheempz, line 82: Does `x.toString(1)` also work like `x.event.toString(1)` where the parameter 1 is afaict a type of formatting flag?

lib/require-patch.js
```diff
@@ -2,7 +2,9 @@
 
 const Module = require('module')
 const path = require('path')
-const ao = require('./index')
+
+const { loggers } = require('./loggers')
```
cheempz, line 6: Curious what the curly braces do?

cheempz (approved):
LGTM with some minor questions.

lib/require-patch.js
ronilan: Object [deconstruction syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)
```
const o = { foo: 'bar' }
const { foo } = o
console.log(foo) // output: bar
```
of what was [exported from `loggers`](https://github.com/appoptics/appoptics-apm-node/blob/master/lib/loggers.js#L114)

lib/log-formatters.js
ronilan: Yes both do. `1` is indeed a formatting flag. [It goes down to the bindings](https://github.com/appoptics/appoptics-bindings-node/blob/master/src/event/event-to-string.cc). There are more formatting options. Test covering them was added [recently](https://github.com/appoptics/appoptics-bindings-node/pull/94). The whole thing is overcomplicated. However, once such an implementation is done and gets "baked in". Removing it becomes a complex refactoring task by itself.

PR #276, opened 2022-06-20
NH-12822-Document-Probe-Configuration-Options

ronilan:
## Overview

This pull adds a basic probe defaults document and links it from the configuration doc.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -36,7 +36,7 @@ The configuration file can supply the following properties, showing their defaul
   insertTraceIdsIntoLogs: false,
   proxy: undefined,
   probes: {
-    // probe-specific defaults. see lib/probe-defaults.js for details
+    // probe-specific defaults. see: /docs/probe-defaults.md for details
   }
 }
 ```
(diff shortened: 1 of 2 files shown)


CONFIGURATION.md
```diff
@@ -61,7 +61,7 @@ The configuration file can supply the following properties, showing their defaul
 ### Configuration File Probe Settings ###
 
 Probes are the packages that `appoptics-apm` auto-instruments. Different types of probes have different
-configuration options. See `lib/probe-defaults.js` for details.
+configuration options. See [probe defaults](./docs/probe-defaults.md) for details.
```
cheempz, line 64: Since we stop mentioning `lib/prob-defaults.js` (as intended since it shouldn't be touched by the customer anymore), we should also update the later paragraph on line 75:
> Probe settings in the `appoptics-apm-config` configuration file will override those in `probe-defaults.js`, so the safest approach to changing an option is to add it to `appoptics-apm-config.json`.

To just something like "customize probe config in appoptics-apm-config" and not mention probe-defaults.js at all.

cheempz (changes requested):
I left a comment, and also noting that this PR still has `appoptics` in the config file name so some rebranding changes will be needed when merged to the solarwinds-apm codebase.

CONFIGURATION.md
ronilan: Reworded in: 92f6f6766f9cc131e939536b023c0920cc9f34d6

Note: Any existing mention of `appopitics` is overwritten by the rebranding commit (assuming `git rebase` works as advertised, and if not, it will not arrive in the other repo. That way of doing thing is almost over. Almost.)

cheempz (approved):
LGTM, thanks for the revisit @ronilan

PR #277, opened 2022-06-23
NH-14027-Fix-q-Probe

ronilan:
## Overview

This pull request fixes issues with the `q` probe so that it only binds when tracing (i.e. in the context of an http request).

## Status

- The probe generated excessive logging when used outside the context of an http request. It that was apparent in test runs and caused customer concern. 
- A fix similar to the one in the pull request was [implemented in the `bluebird` probe](https://github.com/appoptics/appoptics-apm-node/commit/4cbb3a60e36160dbd40e9324223134cc310294d1) a while back.

## Change

Probe was changed to only bind functions when tracing. Tests were cleaned.

## Notes
- Pull Request merges into `nh-main`. Agent built from `master` will stay "unfixed"`.

### DIFF
--- a/lib/probes/q.js
+++ b/lib/probes/q.js
@@ -6,7 +6,9 @@ const ao = require('..')
 function patchThen (proto) {
   if (typeof proto.then !== 'function') return
   shimmer.wrap(proto, 'then', then => function (...args) {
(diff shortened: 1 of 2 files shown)


cheempz (approved):
LGTM

PR #284, opened 2022-07-11
NH-16920-Mask-Logged-Service-Key-when-Invalid

ronilan:
## Overview

This pull request changes logging code so that it masks a service key output in the logged error even if the input value itself is not a valid service key.

## Status

When user is providing an invalid service key, application logs will contain what was provided as-is. An invalid service key may still expose a valid and usable token or part thereof.

## Change

Always mask the service key provided.

## Note

- Pull Request merges into `nh-main`. Agent built from `master` will stay unchanged.

### DIFF
--- a/lib/get-unified-config.js
+++ b/lib/get-unified-config.js
@@ -236,7 +236,7 @@ function getUnifiedConfig () {
     const name = keyParts.join(':')
     const cleansedKey = `${key}:${cleanseServiceName(name)}`
     if (!validKey(cleansedKey)) {
-      fatals.push(`not a valid serviceKey: ${serviceKey}`)
+      fatals.push(`not a valid serviceKey: ${mask(serviceKey)}`)
       results.global.serviceKey = ''
     } else if (cleansedKey !== serviceKey) {
       // the cleansed key is valid but the original key was not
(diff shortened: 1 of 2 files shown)


jerrytfleung:
@ronilan code looks good to me.

@cheempz Is it the idea for agent has its own understanding of service key masking? Although Python, PHP, Node are doing similar masking by `...`. They have different behavior for edge cases.

lib/get-unified-config.js
```diff
@@ -422,14 +422,25 @@ function cleanseServiceName (serviceName) {
   return serviceName.toLowerCase().replace(/ /g, '-').replace(/[^a-z0-9.:_-]/g, '')
 }
 
-// mask the service key so it's not revealed when logging. do not presume a
-// a valid service key, i.e., 64-hex-digits:1-to-255-valid-name
+// mask the service key so it's not revealed when logging.
+// DO NOT presume a valid service key, i.e., 64-hex-digits:1-to-255-valid-name
```
cheempz, line 426: This comment is a bit outdated since in SWO the API Token is typically longer (but I'd also heard talk of generating much shorter ones).  It's probably safest to just assume that the service key needs to have a least 1 char, followed by `:`, followed by at least another char.

test/get-unified-config.test.js
```diff
@@ -537,14 +537,62 @@ describe('get-unified-config', function () {
...
+    })
+
+    it('an invalid serviceKey with no service name is a fatal error (output is masked)', function () {
+      const serviceKey = `${'f'.repeat(32)}`
+      process.env.APPOPTICS_SERVICE_KEY = serviceKey
+
+      const cfg = guc()
+
+      // the serviceKey is not valid and should be reported
+      const fatals = ['not a valid serviceKey: ffff...ffff:']
```
cheempz, line 559: For a service key like `fffffffffffffff`, appending the colon makes the assumption that it's the API Token and that service name is missing.  This is why PHP and Python just mask with something like `ffff...<invalid>`.

cheempz (approved):
LGTM, appreciate the proactive tightening of masking @ronilan.  Per @jerrytfleung's question, yes the main thing is that any possible API Token value is masked and we can leave the details / format to each agent.  I left a few comments that speak to how other agents treat the service key and masking.

PR #285, opened 2022-08-16
Change-Agent-Config-Instructions-to-use-CLI-Require-as-Default

ronilan:
## Overview

This pull request is a documentation change attempting to clarify without complxifiing.

Test is based on what was merged in https://github.com/solarwindscloud/solarwinds-apm-node/pull/15

Live editing on GitHub is welcomed.

## Status
In order to provide instrumentation, the instrumented app has to require the Node agent prior to all other modules.

In the past (AppOptics UI and all docs) users were instructed to add `require('appoptics-apm')` to the top of their entry point JS file.

As long as the instrumented app is “vanilla” JavaScript, those two methods are identical. However, as more and more users use a transpile & build process to create their Node apps, the require (or import) method no longer works since the bundler (a la webpack) is unable to work with the bindings binaries.

While there may be ways to work around it, both at the customer level (using config) and at the agent level (which will require work on the unmaintained `node-pre-gyp`) - just switching to the alternate method should resolve most (or all?) issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,12 +3,8 @@
 The `appoptics-apm` module provides [AppOptics](https://www.appoptics.com/) instrumentation for Node.js.
 
 It supports most commonly used databases, frameworks, and packages automatically. An
-API allows anything to be instrumented.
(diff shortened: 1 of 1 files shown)


tammy-baylis-swi (approved):
Lgtm! I committed one typo fix.

